Issue #1, opened 2020-06-29
Windows Support


ChristophHaag:
For the most part this requires adding a code path for Windows around the OpenGL handling.

Spec: https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_KHR_opengl_enable

In OXRCalls.c we need something like

```c
#ifdef _WIN32
#define XR_USE_PLATFORM_WIN32
#else
#define XR_USE_PLATFORM_XLIB
#endif
```

and a similar ifdef around the code using the `XrGraphicsBindingOpenGLXlibKHR` struct is required, using the `XrGraphicsBindingOpenGLWin32KHR` instead.

This requires getting these values

    hDC is a valid Windows HW device context handle.
    hGLRC is a valid Windows OpenGL rendering context handle.

On Linux the code uses the "current" X display, glx gontext and glx drawable (window)
```c
XOpenDisplay(NULL);
glXGetCurrentContext();
glXGetCurrentDrawable();
```
but ideally we get the gl context and window handles that godot uses from godot itself.

BastiaanOlij:
Half of this is in there  now :)

BastiaanOlij:
I'm going to close this now that we have basic support merged in. Still a long road ahead but we've taken the first major step :)

PR #3, opened 2020-07-05
WIP: Add openxr sdk

BastiaanOlij:
This is just a start bringing the OpenXR SDK in as a submodule and compiling it within the folder structure. Reminded how much I loathe CMake but it is what it is.

I've managed to compile the loader and I'm rejigging the SConstruct file so things will compile. It's not there yet, Windows is doing its usual painful thing that if you don't include windows.h early enough it spits the dummy at you. I'll work some more on it later.

@ChristophHaag , I noticed a lot of the code indentation does not follow the clang-format guidelines Godot enforces so it's being a bit silly with the checking hook.


SConstruct
```diff
@@ -1,83 +1,111 @@
...
+    env.Append(LIBS=["opengl32", "setupapi", "advapi32.lib"])
 
-def add_sources(sources, directory):
-    for file in os.listdir(directory):
-        if file.endswith('.c'):
-            sources.append(directory + '/' + file)
-        elif file.endswith('.cpp'):
-            sources.append(directory + '/' + file)
+elif env['platform'] == "osx":
+    # not tested
```
ChristophHaag, line 60: OpenXR on Mac is not really a thing for now and I suspect the OpenXR SDK itself won't work properly on Mac. For now we only need to concern ourselves with Windows and Linux.

ChristophHaag:
.clang-format is taken from monado and I don't have a particular attachment to it. Feel free to replace it with one from godot and reformat everything.

BastiaanOlij:
> 
 
> .clang-format is taken from monado and I don't have a particular attachment to it. Feel free to replace it with one from godot and reformat everything.

No problem, something for another day when we get things up and running

BastiaanOlij:
So the next thing I'm stuck on is that windows doesn't understand GL extensions. You need to use something like GLAD or GLEW. Lovely. 

@ChristophHaag we're including glad in the openhmd plugins seeing that is what Godot uses internally. See any reason why we shouldn't do the same here?

SConstruct
BastiaanOlij: I've commented out the OSX one and fixed the typo in Linux :)

BastiaanOlij:
@ChristophHaag I did a ton of changes and refactored OVRCalls to be a class now called OpenXRAPI. My initial reason for this is that Microsoft doesn't adhere to newer C standard in their C compiler only adding such changes to their C++ compiler. While it was an improvement MSVC still doesn't like a ton of things. 

You can't use syntax such as `SomeXRStruct elements[some_variable_with_a_count]`, so I had to rewrite those to mallocs and subsequent frees which is UGLY.  Also the struct initializers must have the fields in order, and the error it gives when you do not is total bullocks, tells you the field doesn't exist in the struct. Weird stuff.
And then there is the requirement to always cast malloc... 

One nice thing there is that it did detect some pointer issues in your code, you were allocating structs where you needed to allocate an array of pointers to those structs. Would have ended up allocating far more memory then needed so wouldn't crash but still :)

It all compiles now but it doesn't link yet, it's now complaining about a mismatch in libraries so cmake must have compiled things with different settings. I may go for those precompiled binaries after all :)
I did prepare a bunch of changes where things are different on Windows for Linux, same issues as you ran into not having access to display handles and all that. 

It's late, I'm tired, I'm doing more tomorrow:)
You might want to see if my branch still compiles on Linux, be aware that with these bigger changes it'll be hard to merge anything in that still happens on the older code so please review what I've done so far, the faster we can merge this even if it's not completed the less chance we'll get in eachothers way.

ChristophHaag:
With the changes in https://github.com/BastiaanOlij/godot_openxr/pull/1 the plugin works on linux.

I wrote down what I think is necessary for windows support in #1 btw.

BastiaanOlij:
Cool thanks! So did you compile the loader for Linux as well or pointed it to the pre-installed one in /usr?

Thanks for the issue info, I had already found those and added remarks for them in the code.
I'm going to see tomorrow if I can figure out the compiler errors on Windows and might merge all these changes already and raise a separate PR for finishing making it actually work on Windows. I don't want these bigger changes to stay in an open PR for too long as it holds back anything you want to do.

ChristophHaag:
well, my OpenXR headers are in /usr/include/ and my loader is in /usr/lib, so it just found everything without me having to do anything.
```
$ scons platform=linux target=debug -j32
scons: Reading SConscript files ...
scons: done reading SConscript files.
scons: Building targets ...
g++ -o src/ARVRInterface.os -c -std=c++0x -fPIC -ggdb -O0 -fPIC -I. -Isrc -Igodot_headers -IOpenXR-SDK/include -IOpenXR-SDK/build/linux/include src/ARVRInterface.cpp
g++ -o src/GodotCalls.os -c -std=c++0x -fPIC -ggdb -O0 -fPIC -I. -Isrc -Igodot_headers -IOpenXR-SDK/include -IOpenXR-SDK/build/linux/include src/GodotCalls.cpp
g++ -o src/OpenXRApi.os -c -std=c++0x -fPIC -ggdb -O0 -fPIC -I. -Isrc -Igodot_headers -IOpenXR-SDK/include -IOpenXR-SDK/build/linux/include src/OpenXRApi.cpp
g++ -o src/godot_openxr.os -c -std=c++0x -fPIC -ggdb -O0 -fPIC -I. -Isrc -Igodot_headers -IOpenXR-SDK/include -IOpenXR-SDK/build/linux/include src/godot_openxr.cpp
g++ -o demo/addons/godot-openxr/bin/linux/libgodot_openxr.so -Wl,-R,'$ORIGIN' -shared src/ARVRInterface.os src/GodotCalls.os src/OpenXRApi.os src/godot_openxr.os -LOpenXR-SDK/build/linux/src/loader -lopenxr_loader
scons: done building targets.
```

BastiaanOlij:
@ChristophHaag I'm guessing that is because it only needs to pick up the header files seeing it's no longer in the include path. Or does it add those folders in regardless of what `-I` switches you add?

BastiaanOlij:
Ok, only did a quick change to add a release and debug build of openxr and changed the windows setup to chose the right one. 

Closing in on the final bits, I'm missing the JSON dependency so I have to see where that is obtained from and include it into the setup and for some reason the file system parts aren't linked in. I think this is because openxr is using UCRT and not MSVCRT, not sure if that is because I decided to tell it to compile the loader using makefiles instead of a MSVC project file.

I might just need to look into scripting building the loaded with project files..  Or maybe just use the binaries as @ChristophHaag suggested :)

BastiaanOlij:
I gave up on trying to make the compiled versions work, found out most of my issues were due to missing dependencies and because I'm trying to do the compile through makefiles so it's easier to write instructions in the readme. CMake makes a few weird choices here it seems, it really needs to go through MSVC project files. 

Seeing OpenXR offers prebuild binaries we now include those in the project. It's ashame they can't be submoduled in, I would have preferred that, they are now added in as is. 

I'm going to clean up the readme and then merge the changes so far and start a new PR for windows

PR #5, opened 2020-07-22
Return center position of eyes when requested

### LINKED ISSUE #4: ARVRCamera transform not updated in Godot
Hi, after some talk in the Godot discord, it seems there is a bug with the ARVRCamera Node not being updated properly. It always stays at the origin. (1,0,0 - 0,1,0 - 0,0,1 - 0,0,0)

@BastiaanOlij mentioned that it is only returning the left and right eye positions for rendering, not the center position that is used to track the camera node in Godot. According to him, this is the code that needs to be fixed up: https://github.com/GodotVR/godot_openxr/blob/master/src/ARVRInterface.cpp#L155 

This then results in some buggy behavior, where child nodes don't follow the camera for example (as the parent node transform doesn't seem to change, when it is in fact moving in game). Also, since the transform doesn't update, `ARVRServer.get_hmd_trasnform()` as well as `get_camera_transform()` from the camera node always return the origin (identity rotation and zero translation)

BastiaanOlij:
I can't test this yet because the windows build doesn't work yet but in theory this fixes #4


ChristophHaag:
I implemented this in 883fa0992e615f46d86940304d485643b94e5389 using XR_REFERENCE_SPACE_TYPE_VIEW instead.

PR #6, opened 2020-08-09
Using the new get_native_handle api added to Godot to setup OpenXR for windows

BastiaanOlij:
This PR uses the new get_native_handle api added to Godot in this upstream PR:
https://github.com/godotengine/godot/pull/42531

This API allows us to correctly setup OpenXR for Windows, the first step to getting proper Windows support. 
The same API should be used to improve a few things on the linux side as well but that will come later.

This PR has now been merged so as long as you compile Godot from the 3.2 branch this will work.


BastiaanOlij:
With this done we should now be initialising OpenXR properly for Windows however it is having problems creating the swap chain:
```
Godot Engine v3.2.3.rc.custom_build.3107cb2a7 - https://godotengine.org
OpenGL ES 3.0 Renderer: GeForce RTX 2070 with Max-Q Design/PCIe/SSE2

OpenXR: initialising OpenXR context
OpenXR Runtime only supports OpenGL version 4.3 - 4.6!
Using OpenGL version: 3.3.0 NVIDIA 445.75
Using OpenGL Renderer: GeForce RTX 2070 with Max-Q Design/PCIe/SSE2
Swapchain Formats
805B
881A
881B
8C41
8C43
Using SRGB swapchain!
Failed to create swapchain 0! [XR_ERROR_RUNTIME_FAILURE]
OpenXR init succeeded
```
I wonder if this is because we're running OpenGL 3, not 4 as the error states?

And then crashes:
```
CrashHandlerException: Program crashed
Dumping the backtrace. Please include this when reporting the bug on https://github.com/godotengine/godot/issues
[0] xrNegotiateLoaderRuntimeInterface
[1] xrNegotiateLoaderRuntimeInterface
[2] <couldn't map PC to fn name>
[3] xrDestroyActionSet
[4] godot_openxr_gdnative_init
[5] ARVRInterfaceGDNative::process (C:\Development\godot3-git\modules\gdnative\arvr\arvr_interface_gdnative.cpp:230)
[6] ARVRServer::_process (C:\Development\godot3-git\servers\arvr_server.cpp:358)
[7] VisualServerViewport::draw_viewports (C:\Development\godot3-git\servers\visual\visual_server_viewport.cpp:279)
[8] VisualServerRaster::draw (C:\Development\godot3-git\servers\visual\visual_server_raster.cpp:109)
[9] VisualServerWrapMT::draw (C:\Development\godot3-git\servers\visual\visual_server_wrap_mt.cpp:104)
[10] Main::iteration (C:\Development\godot3-git\main\main.cpp:2123)
[11] OS_Windows::run (C:\Development\godot3-git\platform\windows\os_windows.cpp:3402)
[12] widechar_main (C:\Development\godot3-git\platform\windows\godot_windows.cpp:162)
[13] _main (C:\Development\godot3-git\platform\windows\godot_windows.cpp:184)
[14] main (C:\Development\godot3-git\platform\windows\godot_windows.cpp:196)
[15] __scrt_common_main_seh (d:\A01\_work\6\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288)
[16] BaseThreadInitThunk
-- END OF BACKTRACE --
```

ChristophHaag:
Maybe a gdnative api call for determining godot's preferred texture color format would be useful?

I don't think using OpenGL 3.3 should fundamentally be an issue but it would be possible that the OpenXR runtime or the graphics driver could have a problem with it.

XR_ERROR_RUNTIME_FAILURE at least is not meant to be returned for "expected" errors, but when something else in the runtime goes wrong. Are you running on the Oculus runtime? Perhaps SteamVR beta does something else (or vice versa)?

BastiaanOlij:
Hey @ChristophHaag , 

I'm running the SteamVR beta on this machine, at least it should be running and it does start up steamvr..

Right now the only flavors are GLES2, which requires a GL_RGBA8 texture and GLES3, which requires a GL_RGBA16F texture. Though on GLES3 I think we're still rendering to Godots internal buffer and only the final blit that does all the post effects ends up writing into our texture so even a GL_RGBA8 texture would do there. GLES3 does by default do a linear to SRGB color conversion but you can turn that off and I actually require people to do so for the OpenVR driver. We can query which driver is used through the OS interface in Godot but I don't think we need to because the OpenXR loader returns a valid list of texture formats. You do that SRGB check at the moment but I think it is enough to just pick the first entry (just make sure to turn keep_3d_linear on for your viewport).

So all in all I don't think the texture format itself is the problem here. I think there is something else in the swapchain creation that falls over. No clue yet what though but it does seem to be the only thing left standing in our way to get this puppy running on Windows :)

BastiaanOlij:
Ok, this is working now though there are plenty of things to fix. @ChristophHaag do you think it makes sense to merge the work so far so others can start looking at this as the issues are around controllers and such, it is now rendering the properties.

The bug btw was simply when calling into Godot to get the GL and RC values one of them wasn't calling the function but copying the pointer. Long live casting and bypassing compiler checking...

BastiaanOlij:
@ChristophHaag I've started working on improving the demo a little, can you have a look what it looks like in Linux and whether you're having the same brightness issues as I am?

ChristophHaag:
Sorry for the late reply, was a bit sick.

so I think this looks good for merging except for this

> Note that you will have to compile Godot with that PR included and copy the relevant GDNative files from the godot source into the godot_headers folder or this will not compile.

There currently is no godot_headers repo with these changes so I made one for easy testing without having to compile godot: https://github.com/ChristophHaag/godot_headers. I can confirm that the plugin still works on linux with this PR, the updated headers, and stock godot 3.2.2 without your changes. So I think we should only merge this after your upstream godot PR is approved and/or you make a branch in upstream godot_headers where we can point our submodule at, to keep it possible to compile this plugin without manual copying of files.

Thanks for updating the demo scene, I was thinking about copying the one from godot_openvr git but it looked like quite a bit of work. Off topic but it would be nice to share more assets of the demo scene between vr plugins. :)

Regarding color formats, I noticed you only used `GL_RGBA16F` for windows. On Linux `GL_SRGB8_ALPHA8_EXT` is used on both steamvr and monado. Not sure if the colors are a good to detect this problem but think they look similar enough in godot and in VR, using the GLES3 renderer.
steamvr-srgb-swapchain
monado-srgb-swapchain

BastiaanOlij:
@ChristophHaag that is exactly why I haven't merged it yet and its still marked as WIP. I'm waiting for the upstream changes in Godot to be merged so we can push an update into Godot headers.  Note that you do need to compile Godot to test the Windows version. We'll eventually need to implement similar changes for Linux too. That said, I'm happy with the current workaround for Linux and think our energy is much better spend figuring this out for Vulkan/Godot 4 and look at a better way to obtain this data from the renderer

I plan to do much more with the demo, but that will be separate PRs.

BastiaanOlij:
> Regarding color formats, I noticed you only used `GL_RGBA16F` for windows. On Linux `GL_SRGB8_ALPHA8_EXT` is used on both steamvr and monado. Not sure if the colors are a good to detect this problem but think they look similar enough in godot and in VR, using the GLES3 renderer.

The colors are definitely off for me but the texture format is unrelated to this. This is purely about the color space we're working in. I've had the same issue on OpenVR and talking to the techs over there they only implemented linear color space when 16 bits per color channel are used.

You're using 8 bit per color channel which means OpenVR (and presumably OpenXR over steam) sRGb textures which does ensure the correct color conversion. 

Now with OpenVR I don't have this option because we are using the render buffers Godot uses which have to be 16F and you get banding issues with 8bit but come to think of it, because we're using the external texture approach we can do this for OpenXR so I'll experiment with changing this back.

With the setup we are using for the Oculus SDK and here in OpenXR the GLES3 driver will render the entire scene with internal 16F buffers. The final color grading step that applies DOF, Glow, Tone map, etc. does a copy of the render result into the external texture.

BastiaanOlij:
@ChristophHaag I'm now using the same texture formats as you are and indeed, now it no longer does a double color correction.

I changed the code slightly so we just use the first texture format we can find that meets our needs. OpenXR supposedly sorts these from best to worse. We might need to add more options in at some point in case we run into graphic cards that don't support the two options we prefer.

As a fallback it will always use the first option given to us but as this is often a 16F variant it will have the color issue.
I'm planning to work on a config object like we've done for the other plugins at some point that will allow us to query what decision was made here or give some control over this to the user. We'll see.

For now this works so as soon as our upstream changes in Godot are merged we can merge this and do further changes in new PR.

Owh the only other thing left to discuss is the issues Rune ran into with compiling on Windows. Microsoft really doesn't like some of the newer syntax, I think I got lucky, I must have a slightly newer compiler then Rune was using but if this problem persists I might change a bunch of things to go back to C++17 or something.

src/OpenXRApi.cpp
```diff
@@ -418,34 +432,47 @@ OpenXRApi::OpenXRApi() {
...
+	// We grab the first applicable one we find, OpenXR sorts these from best to worst choice..
 
 	printf("Swapchain Formats\n");
-	for (int i = 0; i < swapchainFormatCount; i++) {
-		printf("%lX\n", swapchainFormats[i]);
+	for (int i = 0; i < swapchainFormatCount && swapchainFormatToUse == 0; i++) {
+		// printf("Found %llX\n", swapchainFormats[i]);
 #ifdef WIN32
-		if (SRGB_SWAPCHAIN && swapchainFormats[i] == GL_SRGB8_ALPHA8) {
+		if (swapchainFormats[i] == GL_SRGB8_ALPHA8) {
```
ChristophHaag, line 449: I I don't know very well what godot does, but I do know when I choose a swapchain with `GL_RGBA8`, then the scene looks too bright here. But for now this works since GL_SRGB8_ALPHA8 is listed first here.

src/OpenXRApi.cpp
```diff
@@ -506,10 +533,10 @@ OpenXRApi::OpenXRApi() {
...
 	glBindTexture(GL_TEXTURE_2D, depthbuffer);
 	glTexImage2D(GL_TEXTURE_2D, 0, GL_DEPTH_COMPONENT24,
-	             configuration_views[0].recommendedImageRectWidth,
-	             configuration_views[0].recommendedImageRectHeight, 0,
-	             GL_DEPTH_COMPONENT, GL_UNSIGNED_INT, 0);
-      */
+		configuration_views[0].recommendedImageRectWidth,
+		configuration_views[0].recommendedImageRectHeight, 0,
+		GL_DEPTH_COMPONENT, GL_UNSIGNED_INT, 0);
+	*/
```
ChristophHaag, line 539: Actually this block can be deleted, just a leftover from the example code I copied.

ChristophHaag (approved):

src/OpenXRApi.cpp
BastiaanOlij: The difference between an RGB and SRGB texture is that OpenGL will do automatic SRGB to Linear conversion on the SRGB texture format. As Godots final result is SRGB (as this is what we want to show on screen) and OpenXR likely expects linear color space as it applies its own color correction for display inside of the headset, having OpenGL internally convert back to linear color space solves the brightness issue.

I don't know however how much overhead is introduced here, it may well be better to use RGB and tell Godot not to do the linear -> SRGB color conversion.

Note also that if GLES2 is used there is no color conversion but all source images will be in SRGB color space anyway

BastiaanOlij: There is also some slight risk of loss of color precision seeing we're using 8bit per color textures here. Though since all the conversions internally are done on floating point values it should be ok.

BastiaanOlij:
Ok, so we've been discussing an alternative to the windows gdnative addition instead adding a method call to the OS singleton that is much more easily enhanced upstream without causing any major breakage. 

Working on making the relevant changes here.

BastiaanOlij:
@ChristophHaag that new commit uses a new approach detailed in #42531, you'll need to build a copy of Godot with that PR included.
With this solution you can use the standard godot_headers as the change is now done through dynamic binding and we're not using godot-cpp. I do want to move some files around but we'll do that after this is merged.

Now that PR needs more work but we can also add the changes needed in the X11 platform as a result to solve your work around. Once we get some feedback that the Godot guys are ok with the PR I suggest we do merge this work in so other work can continue on this.

ChristophHaag:
There is one big problem now and that is the use of `godot_int` for pointers because godot_headers makes that an int `typedef int godot_int;` that is 4 byte here, and pointers are 8 byte here.

With a 64 bit int return type for the get_native_handle function and this code (I think thats the right casts)
```c
	void *display_handle = (void*) os->get_native_handle(OS::DISPLAY_HANDLE, 0);
	void *glxcontext_handle = (void*) os->get_native_handle(OS::OPENGL_CONTEXT, 0);
	void *glxdrawable_handle = (void*) os->get_native_handle(OS::WINDOW_HANDLE, 0);

	graphics_binding_gl.xDisplay = *((Display**) display_handle);
	graphics_binding_gl.glxContext = (GLXContext) glxcontext_handle;
	graphics_binding_gl.glxDrawable = *((GLXDrawable*) glxdrawable_handle);
```
it seems to work fine on monado (but monado doesn't actually rely much on these values).

PR #10, opened 2020-09-14
Skip render when xrLocateViews returns invalid pose

ChristophHaag:
Invalid pose means that the view pose quaternion may not be a valid (normalized) quaternion, which means xrEndFrame() will fail when supplied our projection layer with this view pose.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -9,7 +9,7 @@ opts = Variables(customs, ARGUMENTS)
 env = DefaultEnvironment()
 
 # Define our parameters
-opts.Add(EnumVariable('platform', "Platform", 'windows', ['linux', 'osx', 'windows']))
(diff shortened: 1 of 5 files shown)


BastiaanOlij:
Also important here is if we are retrieving the pose when there is none, we either reuse the last pose or use an identify matrix. Even if we don't have any point in rendering output Godot will still process the frame and could run into trouble processing the scene data. The poses after all do update the camera position and controller positions.

BastiaanOlij:
Ok this now breaks with all the changes I've done, Christoph you mind having a look at it? You might need to redo your changes.

ChristophHaag:
`godot_arvr_get_transform_for_eye()` makes sure to return an identity transform when `get_view_transform()` returns false, there shouldn't be a problem.

BastiaanOlij:
Looks good to me, merge it?

PR #12, opened 2020-10-20
use get_native_handle() for xlib on linux 

ChristophHaag:
On Linux the `get_native_handle()` function in the godot_openxr plugin failed completely and my debugger linked my only to the prototype in godot's OS class https://github.com/godotengine/godot/blob/150f9ce807e7b9be45bf8c030faa80510aea8407/core/os/os.h#L255. No idea what went wrong there but renaming the class in godot_openxr fixed it d7c04f7.


BastiaanOlij:
I'm guessing there is a system OS class in linux? I'm fine with the rename, looks good :)

PR #15, opened 2020-10-30
Added analogue trigger and grip input

BastiaanOlij:
Added support for analog trigger and grip input (and clang format doing its thing on a few unrelated lines)

### DIFF
--- a/src/OpenXRApi.cpp
+++ b/src/OpenXRApi.cpp
@@ -1176,7 +1176,9 @@ void OpenXRApi::update_controllers() {
 
 		// TODO: dynamic binding
 		const int triggerButton = 15;
+		const int triggerAxis = 2;
 		const int grabButton = 2;
+		const int grabAxis = 4;
 		const int menuButton = 1;
 
 #if DEBUG_INPUT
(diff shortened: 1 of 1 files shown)


src/OpenXRApi.cpp
```diff
@@ -1174,10 +1176,13 @@ void OpenXRApi::update_controllers() {
 #endif
 
 		if (triggerStates[i].isActive && triggerStates[i].changedSinceLastSync) {
-			arvr_api->godot_arvr_set_controller_button(godot_controllers[i], triggerButton, triggerStates[i].currentState);
+			float trigger = triggerStates[i].currentState;
+			arvr_api->godot_arvr_set_controller_button(godot_controllers[i], triggerButton, trigger > 0.1);
```
ChristophHaag, line 1180: It would be nice to avoid hardcoded float threshold values like this.

If we want both a float value and a boolean value for the same input, we should make one boolean action and one float action. https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#input-suggested-bindings describes how boolean and float actions behave when they are bound to boolean or float sources.

tl;dr:
1. binding a float action to a boolean input (e.g. oculus go controller only has trigger click, no value): the float will be 0.0 when not pressed and 1.0 when pressed.
2. binding a boolean action to a float input: The runtime will choose a "good" threshold (e.g. 0.5) and the boolean will be true when e.g. pulling the trigger further than the threshold and false when it's less. The runtime may apply some haptic feedback when the threshold is crossed.

For applications that care, XR_VALVE_analog_threshold is an extension that allows modifying the threshold in the 2nd case https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_VALVE_analog_threshold

BastiaanOlij: @ChristophHaag agreed on this, I just needed something to work seeing we don't allow binding yet. It's the same in OpenVR where we leave it up to SteamVR to figure out how to convert the inputs from the controllers to the inputs we require. 

Right now this code is just for backwards compatibility and we require both the binary and analog input for the trigger and grip, even if analog it just switches between 0.0 and 1.0

BastiaanOlij:
This is no longer needed, we implemented the basics of the actions system.

PR #17, opened 2020-11-13
Adding hand tracking support

BastiaanOlij:
This right now is mostly a prototype so not merge able, and also not finished yet. Working on a first version of Hand tracking.

I will likely switch over to using godot-cpp so it becomes much easier to write these things. 

I'll also move the upgrade to loader 1.0.12 into a separate PR.

### DIFF
--- a/src/OpenXRApi.cpp
+++ b/src/OpenXRApi.cpp
@@ -2,9 +2,66 @@
 // Helper calls and singleton container for accessing openxr
 
 #include "OpenXRApi.h"
+#include <math.h>
 #include <OS.hpp>
 
 using namespace godot;
+
+////////////////////////////////////////////////////////////////////////////////
+// Extension functions
+
+XrResult (*xrCreateHandTrackerEXT_ptr)(
+		XrSession session,
+		const XrHandTrackerCreateInfoEXT *createInfo,
+		XrHandTrackerEXT *handTracker) = NULL;
+
+XRAPI_ATTR XrResult XRAPI_CALL xrCreateHandTrackerEXT(
+		XrSession session,
+		const XrHandTrackerCreateInfoEXT *createInfo,
+		XrHandTrackerEXT *handTracker) {
+	XrResult result;
+
+	if (xrCreateHandTrackerEXT_ptr == NULL) {
+		return XR_ERROR_HANDLE_INVALID;
+	}
+
+	return (*xrCreateHandTrackerEXT_ptr)(session, createInfo, handTracker);
+};
+
+XrResult (*xrDestroyHandTrackerEXT_ptr)(
+		XrHandTrackerEXT handTracker) = NULL;
+
+XRAPI_ATTR XrResult XRAPI_CALL xrDestroyHandTrackerEXT(
+		XrHandTrackerEXT handTracker) {
+	XrResult result;
+
+	if (xrDestroyHandTrackerEXT_ptr == NULL) {
+		return XR_ERROR_HANDLE_INVALID;
+	}
+
+	return (*xrDestroyHandTrackerEXT_ptr)(handTracker);
+};
+
+XrResult (*xrLocateHandJointsEXT_ptr)(
+		XrHandTrackerEXT handTracker,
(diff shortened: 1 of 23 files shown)


ChristophHaag:
It looks like this is only for palms so far so I don't have much to say yet.

Apparently I never pushed this wip, here is my unfinished attempt with the glove asset and bones. Not sure if it's of any use. https://github.com/ChristophHaag/godot_openxr/commits/hand_tracking_wip1

BastiaanOlij:
Yup, I did this during my livestream and got to the point of doing this for palms and based it on meshes seeing that was what Rune was asking about. My plan is to do both a mesh and skeleton implementation of this. 

Just converted it to using godot-cpp which cleans this up a lot. One of the main reasons I took time out to add the godot-cpp library.

I see that in your implementation you do a lot more in GDScript. I plan to do as much as I can in C++

BastiaanOlij:
The mesh based approach works now:

image

BastiaanOlij:
Got most of the skeleton implementation working, just need to get good hand models to finish it off with.

BastiaanOlij:
Ok, the version with skeleton support is now working as well. Just two foot notes:

1) waiting on permission to share the meshes we're using
2) the new FBX importer in Godot fails to load the hands correctly, the old one did do it correctly. Either need to wait until that is fixed upstream or see if we can obtain GLTF versions of the files.

ChristophHaag:
On my side this is working really well too https://www.youtube.com/watch?v=a6v6EIcqe0c

This is already quite nice and useful for basic interactions/gestures, we could merge it like this and add more proper hand models later.

BastiaanOlij:
Very cool @ChristophHaag , I'm just waiting on Rune's permission to use the hand assets for the skeleton version before I complete this.

src/OpenXRApi.cpp
```diff
@@ -1208,6 +1339,42 @@ void OpenXRApi::update_controllers() {
...
+			.time = time,
+		};
+
+		result = xrLocateHandJointsEXT(hand_trackers[i].hand_tracker, &locateInfo, &hand_trackers[i].locations);
+		if (xr_result(result, "failed to get tracking for hand %d!", i)) {
+			// For some reason an inactive controller isn't coming back as inactive but has coordinates either as NAN or very large
+			const XrPosef &palm = hand_trackers[i].joint_locations[XR_HAND_JOINT_PALM_EXT].pose;
+			if (
+				!hand_trackers[i].locations.isActive
+				|| isnan(palm.position.x)
```
ChristophHaag, line 1363: > If the returned isActive is true, the runtime must return all joint locations with both XR_SPACE_LOCATION_POSITION_VALID_BIT and XR_SPACE_LOCATION_ORIENTATION_VALID_BIT set.

Pretty sure the position can't be nan and `valid` at the same time.

BastiaanOlij: Yup, that's how it should be, unfortunately that is not how it is, I'm getting junk values when the trackers aren't tracking yet while isActive is true. It's not even NAN all the time, sometimes its just really really really large numbers. 
Either there is something else I need to check to see if the tracking data is valid yet or there is a bug in Valves implementation.

BastiaanOlij:
Seeing I've tested the skeleton implementation with meshes I got from Valve for which I'm waiting for an OK to include them, I suggest we merge this as is even though we don't have hand models to go with them (might see if I can make some simple ones myself as a stand-in. 

What do you reckon @ChristophHaag ?

BastiaanOlij:
Ok, cleaned up all the format and clang issues :)

src/openxr/OpenXRHand.cpp
```diff
@@ -0,0 +1,157 @@
+/////////////////////////////////////////////////////////////////////////////////////
+// Our OpenXR hand GDNative object implemented through meshes
+
+#include <ARVRServer.hpp>
```
BastiaanOlij, line 4: Weird issue to be aware off. If you include openxr before the ARVRServer it gives a really weird nonsensical compile error on Windows.  Hence it is included first, and hence the empty line of clang-format gets upset about the order.

PR #19, opened 2020-11-14
Switching to using godot-cpp

BastiaanOlij:
This PR switches us over to godot-cpp which gives us access to the node wrappers making it far easier to implement more complex logic. 

Because we now rely on logic that is only available in 3.2.4 beta the api.json file in the godot_headers repo is outdated. For the time being I've included a newer one in the root of the project.

Godot-cpp has to be compiled before you compile the plugin using:
Linux: scons platform=linux target=release generate_bindings=yes custom_api_file=../api.json
Windows: scons platform=windows target=release generate_bindings=yes custom_api_file=../api.json


BastiaanOlij:
@ChristophHaag can you have a look if everything keeps working on Linux with these changes?

BastiaanOlij:
I think we need to tweak our clang format checking a bit because it is trying to check files that aren't C/C++ files..

ChristophHaag:
Looks like it is working fine with the Godot 3.2 branch. With godot 3.2.3 `get_native_handle()` segfauts. On linux we can fall back to getting current values if we could find out that we're running on godot <= 3.2.3.

Btw I haven't this, but perhaps such a workaround could be used for windows too https://gitlab.com/madsbuvi/openmw/-/blob/340b4117edd91771768222fb13a2cc21575bc61e/apps/openmw/mwvr/openxrmanagerimpl.cpp#L124

Other than that I'm not a huge fan of having to compile godot-cpp separately from godot_openxr. Could this be integrated into the top level SConstruct?

BastiaanOlij:
Seeing our changes have been merged into the latest 3.2.4 beta and its looking to be released very soon I think we can put up the requirement that this is the minimum build to run OpenXR. 

I could investigate if we can call one scons automatically from another but I'm a little hesitant. The problem with building godot-cpp is that it generates source files from our api.json file which triggers a rebuild of the entire library. That is just wasted time. 
I'd have to find a way to make a switch that triggers this behavior and force that switch to turn on if the files haven't been generated previously.

BastiaanOlij:
@ChristophHaag I haven't figured out a way to make the scons script in godot-cpp run automatically yet, been busy on the hand tracking. 

Are you okay if we merge this in anyway and just raise an issue that in due time we want to improve on compiling this? In the end you only have to compile the godot-cpp library ones and then forget about it, you don't compile it on each run.

ChristophHaag:
Sure, merge whenever it's ready

BastiaanOlij:
Why does Github want to rebase this commit? It's already rebased.. Any ideas if its safe to merge?

BastiaanOlij:
@ChristophHaag mind checking this one more time? I just added namespace so we don't need to prefix everything with godot:: and I change the logging from using printf to Godots internal logging, this ensures you see the info inside of the editor.

BastiaanOlij:
I'm going to work on fixing clang-format tonight or later in the week, i want to add in the logic Ben did in the OpenVR repo so that we actually build the whole repo and get automatic builds people can download and test.

PR #20, opened 2020-11-22
Fixing clang-format on unwanted files and adding automatic builds

BastiaanOlij:
Fixing that clang-format runs on a few files it should not.
And shamelessly stealing @beniwtv work on automatic builds and creating pre-releases (when merged into master only).

**note** at some time in the future when an updated api.json is added upstream we should remove using our custom one from our build script.


BastiaanOlij:
@ChristophHaag I've turned off the check that makes sure the full asset is only assembled on merge, if I've finally got it right it should now build the full artifact. Can you check if it works correctly on Linux?

.github/workflows/build-on-push.yml
```diff
@@ -0,0 +1,152 @@
...
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@v2
+      with:
+        submodules: 'recursive'
+
+    - name: Get OpenXR loader (Ubuntu)
+      run: |
+        sudo add-apt-repository ppa:monado-xr/monado
```
BastiaanOlij, line 34: @ChristophHaag is there a better way then grabbing the loader from Monado?

ChristophHaag (approved):
4.5 megabyte is quite big for the .so but I guess `-g` will do that.

I haven't tested it but it looks good to me.

btw: can the godot_headers submodule be dropped now?

BastiaanOlij:
> btw: can the godot_headers submodule be dropped now?

It should already have been removed as part of merging my godot-cpp PR. It's probably left the folder because of the compiled files.

BastiaanOlij:
> 4.5 megabyte is quite big for the .so but I guess `-g` will do that.

I think Ben was commenting about that the other day too, it's including all the debug symbols.
I think we should fix up those switches, I'll see what I can find out about what he changed in the OpenVR module but if you have suggestions of what the switches should be, let me know.

BastiaanOlij:
@ChristophHaag seeing this does seem to work in its current form, do you think it makes sense to merge this so you can do the tweaking needed on Linux as a new PR?

beniwtv:
On godot-cpp "-g" was removed for release builds, so it probably should be removed here as well if applicable.

BastiaanOlij:
> 
> 
> On godot-cpp "-g" was removed for release builds, so it probably should be removed here as well if applicable.

Ok I've removed the -g setting, lets see how that goes.

Also removed the dlls from the repo.

BastiaanOlij:
I merged this so we can see if the full build works as intended and we can get the other PRs validated, if there are any more tweaks needed I suggest we raise new PRs.

ChristophHaag:
I can confirm that the build from github also works on Linux.

BastiaanOlij:
> 
> 
> I can confirm that the build from github also works on Linux.

Sweet! This is darn cool :), Did the removal of the 'g' switch have an effect on the size?

PR #21, opened 2020-12-24
RFC : Quick addition for thumbsticks support

Miouyouyou:
Greetings,

Since I need to use my Valve Knuckles thumbsticks on a recent Godot OpenXR project of mine, I quickly hacked Knuckles thumbsticks support into this project.

There's clearly a need to build an UI for mapping OpenXR actions to Godot inputs, but that does the trick for the moment.

The thumbsticks are mapped to the Left Joystick on the Left hand and on the Right Joystick on the Right hand... I don't know if that's a good idea, since it's kind of confusing.
People might map the Right Joystick of peripheral 0 (Left hand), instead of peripheral 1 (Right hand) in the Godot editor and start wondering why it's not working correctly.
However, physically, the right hand thumbstick is on the right side of the controller, and left hand thumbstick is on the left side of the controller so, it's "visually correct".

All the mappings have been tested using the Godot Controls editor, meaning :
* creation actions;
* mapping them to joysticks;
* calling Input.get_action_strength("action_name") and  Input.is_action_just_released("action_name");
* using the Valve thumbsticks in-game, through SteamVR on Windows.


BastiaanOlij:
Thanks, I'm hoping to start working on a first version of a UI for mapping actions this month. As an intermediate solution hardcoding this way seems to be ok. @ChristophHaag what do you think?

BastiaanOlij:
Going to start working on proper mappings soon but did some experimenting with this on stream tonight. There is actually a vector input type which is better suited for this.

Miouyouyou:
Indeed, XR_TYPE_ACTION_STATE_VECTOR2F and [XrActionStateVector2f](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/XrActionStateVector2f.html) appear more relevant for sampling joysticks, instead of sampling each vector individually.
It's just that I went into "Monkey see, Monkey copy" way, using some Blender extension code as a basis.

BastiaanOlij:
@Miouyouyou I spend time to finally implement the action system into the plugin so all inputs should now be supported. Let me know if you're happy with how things work now and if it covers everything. We can then close this PR.

Miouyouyou:
Alright, I'll give it a shot this week-end ! Thanks for the heads up !

Miouyouyou:
So, at the moment, on Windows, I tested the latest main branch and the demo worked fine.
My valve knuckles joysticks, A and B button, Trigger and Grab worked without issues !
Nice work !

Miouyouyou:
Since the controls are now working fine, and this pull request code is no longer relevant, I'm closing the pull request.
I'll open a new one if I have new things to add !

PR #23, opened 2021-01-08
vr-common => godot-xr-tools

BastiaanOlij:
Switch from the old vr-common to the new godot-xr-tools


BastiaanOlij:
Annoying formatting rules, need to exclude those for related addons. Will do that over the weekend, bedtime :)

PR #25, opened 2021-01-15
Implementing configurable actions

BastiaanOlij:
I've only just started on this, did a lot of the work so far on stream. Will be updating this description with more information about what I'm building once I'm more sure of the direction. See my stream on 15th of January 2021 for more info for now.

@ChristophHaag I've changed the OpenXRApi class so it can be instantiated without immediately initializing OpenXR. This gives us the ability to start setting configuration options first and then initialize. Have a look at the changes around that if you have time and let me know if you agree with the direction I'm taking with this.

The action stuff itself is too early and not worth looking at yet. I hope to get a long ways during the week to get to a point where it is more clear how it will work. I already went down a few rabbit holes during the stream, I'll probably go through a few more :)


src/openxr/OpenXRApi.cpp
```diff
@@ -403,24 +356,53 @@ OpenXRApi::OpenXRApi() {
...
 		.type = XR_TYPE_SYSTEM_GET_INFO,
 		.next = NULL,
-		.formFactor = XR_FORM_FACTOR_HEAD_MOUNTED_DISPLAY,
+		.formFactor = form_factor,
 	};
 
 	result = xrGetSystem(instance, &systemGetInfo, &systemId);
-	if (!xr_result(result, "Failed to get system for HMD form factor.")) {
-		return;
+	if (!xr_result(result, "Failed to get system for our form factor.")) {
```
ChristophHaag, line 404: By the way: Unlike vulkan, there is a reflection header maintained upstream that can be used to stringify enums.
https://github.com/KhronosGroup/OpenXR-SDK/blob/master/include/openxr/openxr_reflection.h

src/openxr/OpenXRApi.cpp
```diff
@@ -622,7 +625,7 @@ OpenXRApi::OpenXRApi() {
 	// We grab the first applicable one we find, OpenXR sorts these from best to worst choice..
 
 	Godot::print("OpenXR Swapchain Formats");
-	for (int i = 0; i < swapchainFormatCount && swapchainFormatToUse == 0; i++) {
+	for (uint i = 0; i < swapchainFormatCount && swapchainFormatToUse == 0; i++) {
```
ChristophHaag, line 628: can we consistently use `uint32_t` for these?

ChristophHaag (commented):
I haven't run the code but I haven't noticed any big problems. I should watch your live streams. :)

src/openxr/OpenXRApi.cpp
BastiaanOlij: I got warning on Windows when using uint32_t that they were to small I think.. would need to look at that :)

BastiaanOlij:
> 
> 
> I haven't run the code but I haven't noticed any big problems. I should watch your live streams. :)

Yes you should :P I'm a little worried that since this is a larger project where I'm moving things around we're going to have some things clash.

BastiaanOlij:
I've got most of the code in place so we have a memory structure for defining and maintaining actions sets. Need to do some extra work to allow removing actions and to make sure actions are removed from profiles if actions or action sets are removed. 

Also need to change the code that processes the action sets. Still the hardcoded version in there which is now defunct (plugin probably won't run in its current state).

I also want to add some code that converts the configuration to a JSON body or parses a JSON body to create the configuration. That we can use for saving and to provide our default in a much nicer way.

After that I can remove some of the old stuff and start working on making the configuration interactive.

BastiaanOlij:
I've moved some of the support classes I'm writing for actions into separate files. Did get a weird compile error that I still need to look at when I'm sending additional parameters to our xr_result parser. 

Anyway, currently we're still building the action sets and interaction profiles in code, I plan to soon change that over to a configuration file or something.

Currently looking into properly detecting when controllers are discovered so we don't hardcode just the left hand and right hand controllers. 

So currently broken, but getting closer to having a configurable action system. Continuing on this tomorrow.

BastiaanOlij:
I've got most of the code for reacting on actions so that we can support our original controller layer done. Just need to debug it.

I'll work on a JSON parser next so instead of hardcoding building our actionset with instructions, we just have a default json file we use.

Finally we need some objects that allow us to consume user defined actions.

BastiaanOlij:
Pretty much got the logic done to now parse a JSON representation of the action map and implemented a system like I did in OpenVR that if you use the default supplied actions the plugin becomes compatible with the other plugins.

I do still need to complete the default interaction profiles, right now only the simple controller and index controllers are setup.

Now need to add stuff so if someone wants to use their own action map they have the objects to interact with them.

Unfortunately at this point in time there are two major issues I'm stuck on:
1) It sometimes crashes when its creating the actions
2) OpenXR is refusing to apply the interaction profiles, the controllers keep being unbound to our actions

Note that for this reason there is a lot of logging going on that we will be removing.

@ChristophHaag  if you have time to have a look and maybe see if it runs for you, I'd appreciate the second set of eyes on this :)

BastiaanOlij:
Ok, the crash may have been a simple matter of recompiling all again.. *fingers crossed*

BastiaanOlij:
For some reason we can't get this to work properly on SteamVR Windows at the moment, it's not selecting the interaction profile. It also isn't on master anymore which is weird.

So far it works fine on Linux and it works with the Oculus OpenXR runtime.

I'm happy now with how the actions are stored and parsed. Action sets are now stored in a JSON as are the interaction profiles. Both are accessible through the config object so you can change them. 
Changing them however can break the backwards compatibility with ARVRController nodes (expected behavior) but I haven't build the alternative classes yet (See https://github.com/GodotVR/godot-openvr-asset/wiki/OpenVR-actions for how we solved this in OpenVR, I intend to do something similar here).

The demo has been enhanced too, it now shows the state of all the inputs so we can test if they work and I've added teleport and direct movement so you can now navigate around the very empty world.

BastiaanOlij:
@ChristophHaag I need to go through a few of your remarks still but I've changed this to no longer be a draft. I think there is enough in here now that we can merge the logic. There is more to be done but those can be in additional PRs. I think right now we pretty much have a feature match with the other plugins meaning that in theory you could drop this plugin into any Godot VR game that doesn't use platform specific stuff and it will work.

@1runeberg I've left it so the aim pose is used for the controllers as we often pair teleport and pointers to this, you can use the grip pose through an XRPose node (I've got this in the demo)

I've also added a few more things to the demo.

Fixes I still want to do before we merge this PR:
- fix up the file format errors in the static checks
- add some more to the interaction profiles including christophs suggestions
- apply a few of the other suggestions from Christoph such as event->session
- if we can, figure out the cause of the exception error when the app closes (may PR this after this is merged)

For the near future related to this PR:
- Add support for haptics
- Add something that if custom actions are defined you can react to them
- Add support for custom action poses (right now just the hardcoded aim and grip pose actions)
- Add a temporary hand mesh into the demo to show off how the skeleton hand works
- Add a default controller mesh into the demo
- Add better head/body shadow meshes for immersion. 
- Add more interactions to the demo to show off actions

BastiaanOlij:
I've squashed all the commits, did a few cleanup things and added haptics in.
Now for me on my index controller the haptics are really really week, it might be a wrong frequency setting or it might just be that my controllers are misbehaving. 

Looks like the exception error on closing is a regression issue in Godot around unloading the gdnative library too quickly, see
https://github.com/godotengine/godot/issues/45952

1runeberg:
My main feedback on the last build I tested is the need to manually copy to openxrloader to the build directory, not sure if I'm just using scons wrong. It'd be nice and a nice qof for devs if this can be automated.

I didnt notice the crash on exit in the last build I tested, but we do have a known issue with SteamVR crashing when closing an app using an Oculus HMD. We are working on a fix for this.

BastiaanOlij (commented):
> 
> My main feedback on the last build I tested is the need to manually copy to openxrloader to the build directory, not sure if I'm just using scons wrong. It'd be nice and a nice qol for devs if this can be automated.
Should be able to put a copy command in place, the CI does do this when we build the plugin after merging to master.

> 
> I didnt notice the crash on exit in the last build I tested, not sure if related, but we do have a known issue with SteamVR crashing when closing an app while using an Oculus HMD. We are working on a fix for this.

It's a problem introduced in Godot 3.2.4 RC2, you're probably still using RC1 :)

src/openxr/OpenXRApi.cpp
```diff
@@ -51,15 +52,293 @@ XRAPI_ATTR XrResult XRAPI_CALL xrLocateHandJointsEXT(
...
+														   "					\"/user/hand/left/input/trigger/value\",\n"
+														   "					\"/user/hand/right/input/trigger/value\"\n"
+														   "				]\n"
+														   "			},\n"
+														   "			{\n"
+														   "				\"set\": \"godot\",\n"
+														   "				\"action\": \"side_trigger\",\n"
+														   "				\"paths\": [\n"
+														   "					\"/user/hand/left/input/squeeze/value\",\n"
+														   "					\"/user/hand/right/input/squeeze/value\"\n"
```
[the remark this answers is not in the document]
BastiaanOlij: I've added the bindings, not sure if it's enough. I've added this more because it is functionality the other plugins offer as well. Also in godot-xr-tools I use this input to animate the bottom 3 fingers of a hand model.
I agree that the button input is far more useful for the grip.

src/openxr/OpenXRApi.cpp
```diff
@@ -1517,35 +2087,45 @@ void OpenXRApi::process_openxr() {
...
-						Godot::print("OpenXR No interaction profile for {0}", i == 0 ? "/user/hand/left" : "/user/hand/right");
-						continue;
-					}
+					Godot::print("Checking {0} ({1})", inputmaps[i].name, (uint64_t)input_path);
 
-					uint32_t strl;
-					char profile_str[XR_MAX_PATH_LENGTH];
-					res = xrPathToString(instance, prof, XR_MAX_PATH_LENGTH, &strl, profile_str);
-					if (!xr_result(res, "Failed to get interaction profile path str for {0}", i == 0 ? "/user/hand/left" : "/user/hand/right")) {
+					XrResult res = xrGetCurrentInteractionProfile(session, input_path, &state);
```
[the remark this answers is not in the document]
BastiaanOlij: This should now be ok.

Issue #28, opened 2021-01-28
Oculus OpenXR: unable to recover if headset is taken off or moved from face during play session


Nocxr:
Using the Oculus OpenXR runtime If the headset is removed or moved away from the face/head enough that the sensor turns the screen off Godot throws these errors and will not recover until play is restarted:
```
ERROR: OpenXRApi::process_openxr: Abort Mission!
   At: src\OpenXRApi.cpp:1508
ERROR: OpenXRApi::xr_result: OpenXR failed to wait for swapchain image! [XR_ERROR_CALL_ORDER_INVALID]
```
It will continue to spam the XR_ERROR_CALL_ORDER_INVALID line regardless of placing the headset back on, and the HMD screen will rapidly flash blank gray until the play session is ended or restarted.  This is not a huge deal as the initial play session works fine but if your trying to lift the headset up/off to check something on the screen it makes things a little more difficult.

Info:
Windows 10 x64 20H2 19042.746
Radeon Driver: 21.1.1
Godot Version: Godot_v3.2.4-beta6_win64
OpenXR Loader: 1.0.12
Oculus Quest 1 using Oculus Link
Quest Firmware: 1201830.9360.0
Oculus App Version 23.0.0.43.517 (23.0.0.43.517)

HyeokSuLee:
This still problem. 
When i taken off hmd and back, display only shows black screen.
I think need some method to recover session after "OpenXR EVENT: session state changed to XR_SESSION_STATE_STOPPING"

BastiaanOlij:
I could not reproduce this with commit bb85fd21429733d9d4d76319463717f78f7ecdfe on Windows (which bring OpenXR up to 1.0.16). Not sure if that is because some of the changes were already merged before or if this is because I'm using a Rift S on Windows. This may be an issue with the Quest.

HyeokSuLee:
Hi, thanks for your effort.
I'm using lift S too.
So I'v tried with "newest build module ( libgodot_openxr.dll )" + "loader ( openxr_loader.dll -1.0.16 )" but it doesn't show anything in hmd.
Well, "newest build module" + "loader(1.0.12)" works good, but same issue above.

Can i get your loader file? (1.0.16)

BenMcLean:
I have been thinking I should wait to switch my project over to OpenXR until this issue is fixed, but now I'm thinking maybe instead I should switch now to provide more feedback? Or is that needed at this point?

HyeokSuLee:
[...]

PR #29, opened 2021-01-29
don't quit completely on XR_SESSION_STATE_STOPPING

ChristophHaag:
Oculus changes to stopping state when disengaging the proximity sensor.

Monado and apparently SteamVR don't do that (yet).

Not tested on Oculus.

### DIFF
--- a/src/OpenXRApi.cpp
+++ b/src/OpenXRApi.cpp
@@ -1100,6 +1100,11 @@ XrResult OpenXRApi::acquire_image(int eye) {
 }
 
 void OpenXRApi::render_openxr(int eye, uint32_t texid, bool has_external_texture_support) {
+	// no need to do anything if we're stopping or idle
+	if (state == XR_SESSION_STATE_STOPPING || state == XR_SESSION_STATE_IDLE) {
+		return;
+	}
+
 	// printf("Render eye %d texture %d\n", eye, texid);
 	XrResult result;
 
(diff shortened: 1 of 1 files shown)


nocxtious:
Tried this in relation to issue #28, after removing the headset the Godot log shows a slightly different error, there is no abort mission message this time:
```
ERROR: OpenXRApi::xr_result: OpenXR failed to wait for swapchain image! [XR_ERROR_CALL_ORDER_INVALID]

   At: src\OpenXRApi.cpp:137
ERROR: OpenXRApi::xr_result: OpenXR failed to acquire swapchain image! [XR_ERROR_CALL_ORDER_INVALID]
```
and keeps repeating that.  When I put the headset back on instead of rapid flashing gray it is a single frame of the scene (assuming the last one before the headset shuts off) and then a spinning loading hourglass, but still does not seem to recover.

BastiaanOlij:
Yup, it goes nuts! I've managed to add a bit more logic that prevents some of it but as Godot keeps rendering instead of pausing it's not a fix.

BastiaanOlij:
@ChristophHaag same with this one Chris, maybe hold off on merging or it's going to mess up once we merge actions.

BastiaanOlij:
@ChristophHaag something to contemplate here is to use the OpenXRConfig object to send signals back to the Godot side of life to inform the developer a game should be paused or unpaused. while it doesn't stop rendering, it is then possible to implement on the Godot side that the game logic gets paused as well.

BenMcLean:
Might it be possible to put in some kind of temporary solution until the real fix comes along?

It might be acceptable for the Godot app to keep rendering when the headset is off during development, so long as this gets fixed before its time to release the app. It'll drain the battery but only the developer's battery, which is OK for now.

I say this because this issue was reported in January and it is now June. I know the old saying, "Nothing is more permanent than a temporary fix" but even that would be better than just leaving this bug in place.

BastiaanOlij:
@ChristophHaag I think this can be closed now right?

ChristophHaag:
Indeed, it was an incomplete attempt.

PR #31, opened 2021-02-09
scons: Don't set std=c++17 for CCFLAGS

ChristophHaag:
Clang 6.0:

clang -o glad/glad.os -c -fPIC -O3 -std=c++17 -fPIC -Iglad -I. -Isrc -Igodot-cpp/godot_headers -Igodot-cpp/include -Igodot-cpp/include/core -Igodot-cpp/include/gen -Iopenxr_loader_windows/1.0.12/include glad/glad.c
error: invalid argument '-std=c++17' not allowed with 'C'

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -46,9 +46,11 @@ if env['platform'] == "windows":
         env['CC'] = 'clang'
 
         if env['target'] in ('debug', 'd'):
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Seems sound to me :) Merged it if you're happy with it @ChristophHaag

PR #32, opened 2021-02-10
Fixed settings around debugging using vs code

BastiaanOlij:
Adding in debug settings so we can debug this plugin on vs_code


BastiaanOlij:
I also added hooks in for clang-format, need to do something about the other file check too.

PR #34, opened 2021-02-15
Generic Android tech preview

ChristophHaag:
This MR only gets the plugin compiling on the NDK with gradle, it does not integrate into any godot project yet and is untested.

Don't get excitet yet: This MR is just in case people are interested, this will not work on Oculus Quest for the time being. Oculus has their own OpenXR loader stuff, not out of malice but simply because the generic android OpenXR support is still WIP. We have a cardboard-like preview implementation in Monado. More info to that in the readme.


BastiaanOlij:
Dumb question @ChristophHaag  but why did you create a separate android demo? It would be good if we just had one demo that works on all platforms so if there is something stopping us we need to resolve that.

ChristophHaag:
No real reason, just for experiments because I wasn't sure what's really needed. A complete copy of the demo is probably overkill. It's a separate commit, easy to drop.

BastiaanOlij:
Now that I've got most of the Oculus version of this working? what are the differences between that and the official android implementation looking like?

I hope oculus will decide to switch over at some point so we end up with just one android build.

BastiaanOlij:
No need to keep this, Khronos loader support is not part of Godot 4, we currently have no plans to retrofit in Godot 3.

PR #35, opened 2021-02-18
Adding a runtime selection dropdown into the editor

BastiaanOlij:
Adding a dropdown (windows only) to the editor that allows us to select from different (known) runtimes:
image

The known runtimes are stored in `runtimes.json` in our addons folder so we can easily add more entries. On reading the JSON we will check if the paths to the runtime files exist and only offer those we find. We'll probably need to make it smarter to check the same path on different drives in case the user has decided to install Oculus/Steam on a D: or E: drive or something but that is for an enhancement PR.

If the current runtime is not found in the list, it is added (in editor only, the `runtimes.json` is not altered).

### DIFF
--- a/demo/addons/godot-openxr/editor/OpenXRRunSelect.gd
+++ b/demo/addons/godot-openxr/editor/OpenXRRunSelect.gd
@@ -1,0 +1,76 @@
+tool
+extends OptionButton
+
+var available_runtimes : Array = Array()
+onready var platform = OS.get_name()
+
+var home_folder = ''
+
+func _parse_path(p_path):
+	# we might want to add more stuff here at some point
+	p_path = p_path.replace("~", home_folder)
+	return p_path
+
+func _update_tooltip():
+	if selected > 0:
+		var i = get_item_id(selected)
+		hint_tooltip = _parse_path(available_runtimes[i]["path"])
+	else:
+		hint_tooltip = "Select the OpenXR runtime test your project with"
+
+# Called when the node enters the scene tree for the first time.
+func _ready():
+	var current_runtime = 0
+
+	# Parse the users home folder
+	home_folder = OS.get_environment("HOME")
+	if home_folder == '':
+		home_folder = OS.get_environment("HOMEDRIVE") +  OS.get_environment("HOMEPATH")
+
+	# read our json file, may have entries for multiple platforms, we'll filter them later
+	var f = File.new()
+	if (f.open("res://addons/godot-openxr/runtimes.json", File.READ)) == OK:
+		var json = JSON.parse(f.get_as_text())
+		available_runtimes = json.result as Array
+		f.close()
+
+	# check what our current value is
+	var current_path = OS.get_environment("XR_RUNTIME_JSON")
+
+	if available_runtimes.size() > 0:
+		# reset our dropdown if applicable
+		clear()
+		add_item("Default", -1)
+
+		# check which runtimes are actually available
+		var dir = Directory.new()
+		var index = 0
+		for i in available_runtimes.size():
+			var runtime = available_runtimes[i]
+			var path = _parse_path(runtime["path"])
+			if dir.file_exists(path):
+				add_item(runtime["name"], i)
+				index = index + 1
+				if path == current_path:
+					current_runtime = index
+
+		selected = current_runtime
+		_update_tooltip()
+
+		visible = true
+	else:
+		# I guess nothing supported on this platform
+		visible = false
+
+func _on_OpenXRRunSelect_item_selected(index):
+	# this need latest 3.2.4
+	if index == 0:
+		print("Returning to default")
+		OS.set_environment("XR_RUNTIME_JSON", "")
+	else:
+		var i = get_item_id(index)
+		var runtime = _parse_path(available_runtimes[i]["path"])
+		print("Switching to " + runtime)
+		OS.set_environment("XR_RUNTIME_JSON", runtime)
+
+	_update_tooltip()
--- a/demo/addons/godot-openxr/plugin.cfg
+++ b/demo/addons/godot-openxr/plugin.cfg
@@ -1,0 +1,7 @@
+[plugin]
+
+name="Godot OpenXR"
+description="Godot OpenXR plugin"
+author="Christoph Haag and Bastiaan Olij"
+version="1.0.1"
+script="plugin.gd"
--- a/demo/addons/godot-openxr/plugin.gd
+++ b/demo/addons/godot-openxr/plugin.gd
@@ -1,0 +1,14 @@
+tool
+extends EditorPlugin
+
+var openxr_run_select = null
+
+func _enter_tree():
+	openxr_run_select = preload("res://addons/godot-openxr/editor/OpenXRRunSelect.tscn").instance()
+	add_control_to_container(CONTAINER_TOOLBAR, openxr_run_select)
+
+func _exit_tree():
+	if openxr_run_select:
+		remove_control_from_container(EditorPlugin.CONTAINER_TOOLBAR, openxr_run_select)
+		openxr_run_select.queue_free()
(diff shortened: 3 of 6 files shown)


BastiaanOlij:
This sort of works but we don't seem to have enough rights to update the registry key. Looking into that.

BastiaanOlij:
Ok, so I couldn't get changing the registry entry through our plugin to work because of rights issues.

But with a pretty dirty hack I was able to get it to work with `powershell`. 
The biggest issue is that I had to base64 encode the instructions. Powershell does not allow running scripts because somehow that is more unsafe, but it does allow us to base64 encode the script and just pass it through the command line. Go figure.
Problem is that powershell want the source data to be UTF16LE which Godot currently doesn't support as input to the base64 code. I cheated here, it'll work for our usecase. Burvzg added support to Godot 4, we'll need to look into backporting.

demo/addons/godot-openxr/editor/OpenXRRunSelect.gd
```diff
@@ -0,0 +1,86 @@
...
+		# this currently doesn't work due to rights issues
+		# runtime_select.set_runtime(available_runtimes[index]["path"])
+
+		# this does but only in the editor and it's a little hacky
+		var runtime = available_runtimes[index]["path"]
+		var command = 'Set-ItemProperty -Path \'HKLM:\\SOFTWARE\\Khronos\\OpenXR\\1\' -Name \'ActiveRuntime\' -Value \'' + runtime + '\''
+		# var command = "Write-Host 'Test'"
+
+		# powershell wants unicode16, for now this is a DIRTY work around
+		# utf16 support is coming
```
BastiaanOlij, line 72: This is the only part I'm not happy about but for now I don't think we can do anything but accept it, unless someone wants to write a UTF-8 to UTF16LE parser in GDScript.
I'm keeping an eye out for Bruvzg work and whether a utf16 option becomes available in GDScript or I might add something myself in the not to distant future. Either way I suggest we handle this in a future PR.

demo/addons/godot-openxr/plugin.cfg
```diff
@@ -0,0 +1,7 @@
+[plugin]
+
```
BastiaanOlij, line 2: Note, adding the plugin section allows us to add editor based stuff, disabling the plugin here only disables the editor enhancements not the actual plugin. We could have some more fun with this in due time.

demo/addons/godot-openxr/runtimes.json
```diff
@@ -0,0 +1,14 @@
...
+    {
+        "name": "SteamVR",
+        "path": "C:\\Program Files (x86)\\Steam\\steamapps\\common\\SteamVR\\steamxr_win64.json"
+    },
+    {
+        "name": "Oculus",
+        "path": "C:\\Program Files\\Oculus\\Support\\oculus-runtime\\oculus_openxr_64.json"
+    },
+    {
+        "name": "Microsoft MR",
```
BastiaanOlij, line 11: I've added Microsoft to the list even though we don't support it right now due to lack of DirectX support. Still someone might want to be able to change it back to MS before leaving Godot.

src/openxr/OpenXRRuntimeSelect.cpp
```diff
@@ -0,0 +1,93 @@
...
+	return return_value;
+#else
+	Godot::print("OpenXRRuntimeSelect.get_runtime(): Unsupported platform");
+	return String("n/a");
+#endif
+}
+
+bool OpenXRRuntimeSelect::set_runtime(const String p_runtime) {
+#ifdef WIN32
+	// TODO possibly rewrite to use RegOpenKeyExW/RegSetValueExW
```
BastiaanOlij, line 69: Should have removed this comment :P 

I'm leaving this function here even though it will fail due to rights issues. Possibly in the future we should investigate if we can elevate the plugin at least in the editor.

demo/addons/godot-openxr/runtimes.json
```diff
@@ -0,0 +1,22 @@
+[
+    {
+        "name": "Monado",
+        "path": "/usr/share/openxr/1/openxr_monado.json"
+    },
+    {
+        "name": "SteamVR",
+        "path": "~/.steam/steam/steamapps/common/SteamVR/steamxr_linux64.json"
+    },
```
BastiaanOlij, line 9: @ChristophHaag I have no idea if the ~ works, we may need to add some code that it replaces it with the users home folder.

BastiaanOlij:
@ChristophHaag I've completely changed this to be based on the environment variable.
It needs: https://github.com/godotengine/godot/pull/46413
Should in theory work on Linux but as I can't test I don't know, it may have a problem with the ~ not being parsed to the current user folder.

demo/addons/godot-openxr/runtimes.json
ChristophHaag: It does not work, and kinf of expected, `$HOME` also isn't automatically resolved.

This would be the easy fix for linux
```
var home = OS.get_environment("HOME")
...
var path = runtime["path"].replace("~", home)
```

ChristophHaag: same handling will be needed in _on_OpenXRRunSelect_item_selected

demo/addons/godot-openxr/runtimes.json
ChristophHaag: Switching back isn't an issue anymore but I would still leave it in because why not.

src/godot_openxr.cpp
```diff
@@ -5,6 +5,7 @@
 // with loads of help from Thomas "Karroffel" Herzog
 
 #include "godot_openxr.h"
+
```
ChristophHaag, line 8: empty newline :)

demo/addons/godot-openxr/plugin.gd
```diff
@@ -0,0 +1,15 @@
+tool
+extends EditorPlugin
+
+var openxr_run_select = null
+
+func _enter_tree():
+	# multiple runtimes are currently only supported on Windows
```
ChristophHaag, line 7: This comment is now wrong.

demo/addons/godot-openxr/plugin.cfg
```diff
@@ -0,0 +1,7 @@
...
+name="Godot OpenXR"
+description="Godot OpenXR plugin"
+author="Christoph Haagch and Bastiaan Olij"
```
ChristophHaag, line 5: The name is Christoph Haag, ch are the initials :)

ChristophHaag (commented):
With the home directory fix it works nicely on linux.

Short wishlist (maybe for the future):
* Show the full path somewhere in the ui (tooltip on the dropdownbox?)
* Persistently store the selected runtime
* Allow adding runtimes from the UI

demo/addons/godot-openxr/plugin.cfg
BastiaanOlij: I never knew that :)

demo/addons/godot-openxr/runtimes.json
BastiaanOlij: Ok it's HOMEDRIVE and HOMEPATH on windows, so I'll add that in just for completeness sake.
I'm putting the logic to replace the home path into a function so we can extend it later on. Also keeping the path as is inside of the array so we can add items and save it back to a json at some point.

BastiaanOlij:
> Short wishlist (maybe for the future):

I've added the tooltip but the other two I think we should do in new PRs.
One thing to discuss here is where we store the updated JSON. I consider the addons folder in a project to be read only because you need to be able to replace it with an updated version when we release a new version of the plugin.

BastiaanOlij:
@ChristophHaag if you can have another quick look, I think this is ready to be merged. It works really well imho

ChristophHaag:
looks good to me

PR #40, opened 2021-05-15
Change OpenXR loader from 1.0.12 to 1.0.16

BastiaanOlij:
Updating the windows loader to 1.0.16. Have to do some more testing to make sure if everything still works, will do that on Monday.

@ChristophHaag do we need to change anything on the Linux side for this?


ChristophHaag:
None of this is used on linux so if the CI is happy, it should be okay.

PR #42, opened 2021-05-27
add support for compile_commands.json

paul-marechal:
The clangd language server uses compile_commands.json to provide rich
IDE support. This requires SCons version >=4.0

See https://scons.org/doc/production/HTML/scons-user/ch27.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ logs
 .vscode
 *.pdb
 *.ilk
+compile_commands.json
+.cache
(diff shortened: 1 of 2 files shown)


paul-marechal:
If there is a way to do that without this commit please tell me :)

BastiaanOlij:
Sorry for the late response, been a little busy. Hmm I need to see if I can safely upgrade to scons 4, I'm still on 3. Is it possible to add a check in so we don't loose compatibility with scons 3?

paul-marechal:
I didn't find a way to test the SCons version from the script, so I try/except the `env.Tool` call instead...

BastiaanOlij:
Let's give it a try and see what people say

PR #43, opened 2021-06-16
Updated contributors, it was still the original one from OpenHMD

BastiaanOlij:
We still have the original contributors file from the OpenHMD plugin. I don't think the current plugin has much in common with that anymore :) 

Updated it to reflect who have been contributing here.

That said, Christoph let me know if I should leave some of the names in there if they contributed to the original plugin before we moved it here.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,11 +1,10 @@
 Contributors
 ============
 
-This plugin has mostly been developed by [Christoph Haag](https://github.com/ChristophHaag/).
+This plugin has mostly been developed by:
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
Also added @beniwtv in, he's not added any PRs but all the workflow logic was created by him for the OpenVR plugin so it makes sense to credit him in the contributors file.

@ChristophHaag let me know who else we should put (back) in from when you hosted the plugin on your git repo?

ChristophHaag:
Just me.

beniwtv:
I still plan to add a PR for overlay support here, when I get around to it :)

PR #46, opened 2021-06-22
Add workflow logic to create artifact on release

### LINKED ISSUE #44: Artifact not created on release
Need to update workflow actions so artifacts are also created on releases...

BastiaanOlij:
Thanks to @beniwtv added logic to also create artifacts on release.

Fixes #44


BastiaanOlij:
Won't know 100% for sure if this works until we merge and tag so here it goes....

PR #47, opened 2021-06-22
Fix re-init of action sets when headset refocusses

BastiaanOlij:
Finally got around to testing OpenXR with Oculus Link and indeed got things to crash when you take off the headset and put it back on.

Oculus Link gets an `XR_SESSION_STATE_FOCUSED` every time you put the headset back on and we were initializing the action set at this point in time. As it already was initialized the second initialization failed.

I'm keeping a status now to prevent double init. There is probably more to this in the future if we ever have a reason for re-initializing the action set but that's for another day.

### DIFF
--- a/demo/addons/godot-openxr/CHANGES.md
+++ b/demo/addons/godot-openxr/CHANGES.md
@@ -1,6 +1,10 @@
 Changes to the Godot OpenXR asset
 =================================
 
+1.0.1
+-------------------
+- Fix crash issue on Oculus Link when taking headset off and putting it back on.
+
 1.0.0
 -------------------
 - Original implementation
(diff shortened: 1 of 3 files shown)


BastiaanOlij:
@ChristophHaag as discussed I haven't got any more clarity on the issue with Oculus Link changing the session status to `XR_SESSION_STATE_STOPPING`, this PR does fix the other issue around action sets. 
Also I changed it so that we don't call `xrBeginFrame` nor submit nor call `xrEndFrame` if `shouldRender` is false, the OpenXR docs seem to indicate that is important. 

Let me know what you think, I think we should merge this as is and then do a follow up PR if we get some answers around the Oculus Link issue.

BastiaanOlij:
I'm going to merge the work so far as it seems to work, and then create a new PR for figuring out the lifecycle issues.

PR #48, opened 2021-06-23
Add motion range support

BastiaanOlij:
This implements the motion range extension provided this is supported by the OpenXR runtime.

This allows you to toggle the range for finger tracking:
- Unobstructed will make a closed first when the user grabs the controller (default)
- Conform to controller will limit the range of the fingers so it grabs the controller 
These are the only two models currently supported by OpenXR but more will likely be added in the future. 

You can switch between the modes at runtime, so you could use unobstructed and not show the controller, then when something is picked up you can go back to the conform to controller mode. 

Note that there is only one setting tracked per hand so if you use multiple Hand/Skeleton nodes for each hand the last one wins. It's assumed you don't represent a hand more then once.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -74,6 +74,9 @@ elif env['platform'] == "linux":
     target_path += "linux/"
     godot_cpp_library += '.linux'
 
+    # make sure we have access to the correct headers, cheating here a little.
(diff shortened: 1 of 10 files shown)


BastiaanOlij:
I'll merge this after the other 1.0.1 PR is merged so I can update the changes document.

PR #49, opened 2021-06-25
Reworking lifecycle

BastiaanOlij:
Just a quick and dirty first pass so far but working on moving parts of the initialisation of the plugin to the proper timing as far as the OpenXR lifecycle is concerned.

Right now we do almost all initialisation right away but according to the documentation:
https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#session

We're supposed to wait on state changes once the session is created before proceeding. 

This should also enable us to react properly on the OpenXR runtime requesting a session to be ended and on quitting the application.

### DIFF
--- a/src/openxr/OpenXRApi.cpp
+++ b/src/openxr/OpenXRApi.cpp
@@ -748,13 +748,12 @@ bool OpenXRApi::initialiseSession() {
 
 	// TODO We should add a setting to our config whether we want stereo support and check that here.
 
-	XrViewConfigurationType viewConfigType = XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO;
-	if (!isViewConfigSupported(viewConfigType, systemId)) {
+	if (!isViewConfigSupported(view_config_type, systemId)) {
 		Godot::print_error("OpenXR Stereo View Configuration not supported!", __FUNCTION__, __FILE__, __LINE__);
 		return false;
 	}
 
-	result = xrEnumerateViewConfigurationViews(instance, systemId, viewConfigType, 0, &view_count, NULL);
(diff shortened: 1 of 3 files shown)


BastiaanOlij:
So this works on SteamVR but on Oculus runtime it gets stuck, the log shows:
```
OpenXR increased use count to 6
OpenXR initialiseInstance
OpenXR initialiseExtensions
OpenXR initialiseSession
OpenXR Runtime only supports OpenGL version 4.0 - 4.6!
OpenXR Using OpenGL version: 3.3.0 NVIDIA 466.63
OpenXR Using OpenGL renderer: NVIDIA GeForce RTX 2070 with Max-Q Design/PCIe/SSE2
OpenXR Interface initialized
Deactivated LeftHandController
Deactivated RightHandController
OpenXR EVENT: session state changed to XR_SESSION_STATE_IDLE
OpenXR EVENT: session state changed to XR_SESSION_STATE_READY
OpenXR initialiseSpaces
OpenXR initialiseSwapChains
OpenXR Swapchain Formats
OpenXR Using RGBA swapchain.
OpenXR View pose became valid
OpenVR Head pose is now valid
```
And it just hangs on calling xrWaitFrame, so it looks like it doesn't manage to synchronize. Still looking into why...

BastiaanOlij:
Ok, I undid the changes I made around `shouldRender` and that made it all work. There are issues with things leaking at the moment and with action sets probably failing, so I'll leave this on draft, I'll look into that some more tomorrow but I think we're close to this working properly.

BastiaanOlij:
This is now working correctly as far as I can tell. It nicely cleans up when the state changes to session_stopping, and picks up things nicely again when we get session_ready.

I've moved loading the actions sets into the initialisation step but then moved the binding/unbinding to resp session_ready/session_stopping.

PR #50, opened 2021-06-29
Fix folder structure of godot_openxr.zip created by Github actions

BastiaanOlij:
The asset library bases deployment on zip files produced from a github repository which start with a folder name based on the repos name. 
This causes the `addons` folder in the zipfile to be ignored, we need to add a parent folder


BastiaanOlij:
I guess there is no way to test this but go....

PR #56, opened 2021-08-03
Add building openxr plugin for quest

BastiaanOlij:
Adding gradlew configuration to compile the plugin for Android currently including Oculus' OpenXR loader

This does NOT work yet but I think we're close, everything compiles and I think I have all the code changes in there that are needed.


BastiaanOlij:
Refer to https://developer.oculus.com/downloads/package/oculus-openxr-mobile-sdk for extra info but all thats needed now is to actually setup the plugin so we can properly export it. This involves telling Android to start the OpenXR loader before running Godot and provide a few flags in the manifest file.

BastiaanOlij:
Also must not forget to attempt setting up CI so we can build our deliverable

PR #58, opened 2021-08-04
Android dev

BastiaanOlij:
For Android/Quest development, raise PRs against the android-dev branch.

This PR is to merge it all into master once we're done.


demo/addons/godot-openxr/godot_openxr.gdnlib
```diff
@@ -9,10 +9,10 @@ reloadable=false
 
 Windows.64="res://addons/godot-openxr/bin/win64/libgodot_openxr.dll"
 X11.64="res://addons/godot-openxr/bin/linux/libgodot_openxr.so"
-Android.64="res://addons/godot-openxr/bin/android/libgodot_openxr.aar"
+Android.64="res://addons/godot-openxr/bin/android/libgodot_openxr.so"
```
BastiaanOlij, line 13: @m4gr3d Like we discussed on chat, I believe this should be .aar so the Godot export knows that this is the file that it needs to process, but during export we look into a way that the runtime knows the aar file extracts an so file and puts it in this place.

BastiaanOlij: This will require changes in Godot, so while we're still doing this the same way as with the quest plugin we'll leave this as `.so`

BastiaanOlij: After talking this through and further discussing the solution Fredia came up with, very happy with the end result.

BastiaanOlij:
@ChristophHaag @m4gr3d The only thing not working at the moment is finger tracking but we can live without that for now. 
It also requires https://github.com/godotengine/godot/pull/51780 to get the correct colors but that has now been approved and will likely be merged. 

That does mean we can't release `1.1.0` until Godot 3.4 is out but that will happen soon after Akien returns from holidays.

I suggest we merge this branch into master. I've created a `1.0.x` branch for the current version so we can apply patches for fixes to maintain support for Godot 3.3

Let me know what you guys think. I'm planning to start work on the proper Godot 4 port later in the week now that most of the GD extension logic has been merged (the replacement for the old GDNative layer).

m4gr3d:
Merging into `master` sounds good!
Is there an issue that tracks the hand tracking bug?
Also, can you give push/write access to the repo.

m4gr3d (approved):

PR #62, opened 2021-08-05
Change demo for Android

BastiaanOlij:
Making adjustments to the demo so it will run on Android

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 *.os
 *.so
 *.dylib
-.import
 *.dblite
 *.exp
 *.lib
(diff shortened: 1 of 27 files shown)


.gitignore
```diff
@@ -32,7 +31,3 @@ local.properties
 build
 
 # Godot metadata files
-*.import
```
BastiaanOlij, line 35: Moved these into an ignore file in the demo folder itself, I am probably going to spend some more time moving all the ignore filters specific to a Godot project there. Makes it easier to use the demo stand alone, thinking about adding a second downloadable asset with a completely working demo project.

PR #64, opened 2021-08-05
Complete setup of the Quest OpenXR Godot plugin

m4gr3d:
To run on the Quest:
- generate the shared libs
- update the addons directory with them
- disable custom build if it's selected
- enable the usual vr options (e.g: Oculus mobile VR, 6dof, hand tracking optional, focus awareness)

Then build and run on the Quest


m4gr3d:
cc @BastiaanOlij

BastiaanOlij (approved):
As we discussed on chat, I'm not entirely convinced of the need of moving the `gdnlib` and `gdns` files into their own folder. It goes against established conventions for Godot plugins but it can't do much harm I guess. To keep momentum going we'll keep this change.

Do note that I'm very concerned if this is changed in the same way that the Oculus Quest plugin works, with these files moved into the AAR instead of remaining in the plugin folder we'll get into trouble with the other platforms and with editing the scenes. 

We under utilise this in the Quest plugin because without a Windows/Linux binary that functionality is broken but the editor uses the `gdnlib` and `gdns` files to load the plugins within the editor and update its meta data and utilise any of the toolclasses found within the libraries.

PR #70, opened 2021-08-12
Rework ARVRInterface so we inherit, requires 3.4

### LINKED ISSUE #53: Refactor ARVRInterface to subclass ARVRInterfaceGDNative
This is something we've only applied to the (still in development) ARCore plugin but we added a few changes to the GDNative interface that allows us to subclass the `ARVRInterfaceGDNative` wrapper and create more natural additions to the interface.

Combined with the planned improvements in GDNative for Godot 4 this will help developers lots.

BastiaanOlij:
This reworks our `ARVRInterface` so we nicely inherit from `ARVRInterfaceGDNative`. 

While many of the internal functions aren't actually exposed through GDNative implementing it this way allows us to more easily add methods and properties to the class that are actually exposed to Godot. We're still using the interface struct to handle all the internal functions but these have now become pass-troughs.

In theory this should make a move to Godot 4.x easier as well.

Waiting for an official 3.4 release and `godot-cpp` being updated before this is ready for merging (CI will also fail until `godot-cpp` is updated).

Fixes #53


BastiaanOlij:
Ok, rebased this and did some rework on this just in case we do decide to go ahead with this. At this point in time I think it's not worth it but there are one or two features that do become nicer if we do.
You do need a recent 3.4 build and update the `api.json` file. Note that at the time of writing this, there are a few mistakes in the generated `api.json` file. Nothing that can't be fixed by one or two edits, but annoying none the less

I've renamed the class to XRInterfaceOpenXR as to make the transition to Godot 4 more sensible. 

The main issue right now is this:
```
void GDN_EXPORT godot_openxr_gdnative_singleton() {
	godot::ARVRServer::get_singleton()->add_interface(godot::XRInterfaceOpenXR::get_singleton_instance());
}
```

It looks like GDNative's singleton initialisation logic is run too early. While the ARVRServer is up and running, GDNative uses the singleton registration in ClassDB which is setup later in the startup code. This code thus fails to obtain the ARVRServer singleton instance...

BastiaanOlij:
No point in changing this over.

PR #73, opened 2021-08-16
Add a `joystick_button` action.

Zoomulator:
I took the liberty of adding thumbstick clicking to the default actions and interactions since the configurable actions may take a while. :)

I chose button index `14` by referring to the JoystickList enum:
```JOY_VR_PAD = 14 --- Push down on the touchpad or main joystick on a VR controller.```

I'm not entirely sure it's the correct index, since the valve knuckles has a pad and main joystick, so which should use that index really?
I would not object if you'd like another button index instead!


BastiaanOlij:
Looks good, make sure to update the `demo/addons/godot-openxr/CHANGES.md` file and add your name to `CONTRIBUTORS.md` and I'll merge it

Zoomulator:
@BastiaanOlij Cool! All done :)

BastiaanOlij:
Thanks!

PR #79, opened 2021-08-21
Fix occurrence of infinity transform from pose.

Zoomulator:
OpenXR may give a zero length quaternion at initialization causing a
transform with inf values, resulting in a slew of assertion errors from
`VisualServer.instance_set_transform` if a VisualInstance happens to be
a child of OpenXRHand, OpenXRPose or OpenXRSkeleton.

The OpenXR specification states that the XrQuaterionf should be of unit
length, so I figured `is_normalized` is a good health check.

### DIFF
--- a/demo/addons/godot-openxr/CHANGES.md
+++ b/demo/addons/godot-openxr/CHANGES.md
@@ -1,6 +1,10 @@
 Changes to the Godot OpenXR asset
 =================================
 
+1.0.4
+-------------------
+- Fix occurrence of invalid transform from pose.
+
 1.0.3
 -------------------
 - Copy loader dll in place when compiling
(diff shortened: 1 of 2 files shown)


BastiaanOlij (approved):
Simple but effective fix :) Will merge tomorrow unless someone has more feedback.

ChristophHaag:
do you happen to know if the `XR_SPACE_LOCATION_ORIENTATION_VALID_BIT` is set when this happens? And on what runtime/device is that?

Zoomulator:
@ChristophHaag I wasn't aware of its existence! :D I'll take a look.

Zoomulator:
@ChristophHaag @BastiaanOlij 
OK, so on closer inspection it seems that Action and OpenXRHand actually checks for the valid flags on the `XrSpaceLocation` before using the `transform_from_pose`, the OpenXRPose and OpentXRSkeleton does not.
Should I make a `transform_from_space_location` they all can use that validates the space first instead?

Zoomulator:
Oh, also, Action only checks for the orientation valid bit and has the location valid check commented out. Do you know the reason?

Zoomulator:
This is a really bad fix :D
Noticed that more than just the zero-length quaternions were caught.
I'll close this.

ChristophHaag:
> Oh, also, Action only checks for the orientation valid bit and has the location valid check commented out. Do you know the reason?

I wasn't super sure how 3dof devices are going to do this. Monado now has hardcoded "valid" positions for 3dof devices, not sure if everyone has.

PR #80, opened 2021-08-21
Validate space locations in OpenXRSkeleton and OpenXRPose.

Zoomulator:
OpenXR may give a zero length quaternion at initialization causing a
transform with inf values, resulting in a slew of assertion errors from
VisualServer.instance_set_transform if a VisualInstance happens to be
a child of OpenXRPose or OpenXRSkeleton.

This is a more rigorous fix than the one I first proposed: https://github.com/GodotVR/godot_openxr/pull/79

### DIFF
--- a/src/openxr/OpenXRApi.cpp
+++ b/src/openxr/OpenXRApi.cpp
@@ -2742,3 +2742,26 @@ Transform OpenXRApi::transform_from_pose(const XrPosef &p_pose, float p_world_sc
 
 	return Transform(basis, origin);
 }
+
+template <typename T>
+Transform _transform_from_space_location(OpenXRApi &api, const T &p_location, float p_world_scale) {
+	Basis basis;
+	Vector3 origin;
+	const auto &pose = p_location.pose;
+	if (p_location.locationFlags & XR_SPACE_LOCATION_ORIENTATION_VALID_BIT) {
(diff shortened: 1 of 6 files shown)


src/openxr/OpenXRApi.cpp
```diff
@@ -2590,3 +2590,21 @@ Transform OpenXRApi::transform_from_pose(const XrPosef &p_pose, float p_world_sc
 
 	return Transform(basis, origin);
 }
+
+template <typename T>
+Transform _transform_from_space_location(OpenXRApi &api, const T &p_location, float p_world_scale) {
+	const auto flags = XR_SPACE_LOCATION_ORIENTATION_VALID_BIT | XR_SPACE_LOCATION_POSITION_VALID_BIT;
```
BastiaanOlij, line 2596: I think we should break this up. Start with an identify Transform, then:
- if `XR_SPACE_LOCATION_ORIENTATION_VALID_BIT` is set we convert the quarternion to a basis matrix
- if 'XR_SPACE_LOCATION_POSITION_VALID_BIT' is set we assign the origin

That would be the most accurate way of doing this.

Zoomulator:
Makes sense! I'll do that later today 😄

Zoomulator:
@BastiaanOlij All done!

BastiaanOlij:
Thanks!

PR #83, opened 2021-08-31
Update to version 1.0.18 of the Oculus OpenXR loader.

m4gr3d:


BastiaanOlij (approved):
Sweet, now we're in line with the Khronos one again :)

PR #84, opened 2021-09-04
Misc code cleanup

m4gr3d:
Code cleanup, and fix misc bugs:
- XR_ERROR_CALL_ORDER_INVALID when the xrReleaseSwapchainImage was called after a call to xrWaitSwapchainImage had timed out.
- Fix proper pause/resume lifecycle on Oculus Quest devices.

### DIFF
--- a/src/ARVRInterface.cpp
+++ b/src/ARVRInterface.cpp
@@ -2,6 +2,7 @@
 // Our main ARVRInterface code for our OpenXR GDNative module
 
 #include "ARVRInterface.h"
+#include <MainLoop.hpp>
 
 typedef struct arvr_data_struct {
 	OpenXRApi *openxr_api;
@@ -23,7 +24,7 @@ godot_int godot_arvr_get_capabilities(const void *p_data) {
 	godot_int ret;
 	ret = 2 + 8; // 2 = ARVR_STEREO, 8 = ARVR_EXTERNAL
 	return ret;
-};
+}
 
 godot_bool godot_arvr_get_anchor_detection_is_enabled(const void *p_data) {
 	godot_bool ret;
@@ -31,11 +32,11 @@ godot_bool godot_arvr_get_anchor_detection_is_enabled(const void *p_data) {
 	ret = false; // does not apply here
 
 	return ret;
-};
+}
 
-void godot_arvr_set_anchor_detection_is_enabled(void *p_data, bool p_enable){
+void godot_arvr_set_anchor_detection_is_enabled(void *p_data, bool p_enable) {
 	// we ignore this, not supported in this interface!
-};
+}
 
 godot_bool godot_arvr_is_stereo(const void *p_data) {
 	godot_bool ret;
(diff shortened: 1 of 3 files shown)


src/ARVRInterface.cpp
```diff
@@ -289,6 +287,22 @@ int godot_arvr_get_external_texture_for_eye(void *p_data, int p_eye) {
...
 void godot_arvr_notification(void *p_data, int p_what) {
 	// nothing to do here for now but we should implement this.
+	auto *arvr_data = (arvr_data_struct *)p_data;
+	if (arvr_data->openxr_api == nullptr) {
+	  return;
+	}
+
+	switch (p_what) {
+      case 1014: // NOTIFICATION_APP_RESUMED - The Android activity is resumed.
+        arvr_data->openxr_api->on_resume();
```
BastiaanOlij, line 297: we should have these enums somewhere in godot-cpp?

src/openxr/OpenXRApi.cpp
```diff
@@ -631,34 +672,44 @@ bool OpenXRApi::initialiseInstance() {
...
-		Godot::print_error("OpenXR Runtime does not support OpenGLES extension!", __FUNCTION__, __FILE__, __LINE__);
-		free(extensionProperties);
-		return false;
-	}
-
-	if (!isExtensionSupported(XR_KHR_ANDROID_CREATE_INSTANCE_EXTENSION_NAME, extensionProperties, extensionCount)) {
-		Godot::print_error("OpenXR Runtime does not support android instance extension!", __FUNCTION__, __FILE__, __LINE__);
-		free(extensionProperties);
-		return false;
+	// Check that the extensions required are present.
```
BastiaanOlij, line 691: Hmm, this leaves me wondering if we shouldn't build a vector with required extensions, then populate them according to platform or other switches, and have the rest of the logic the same for all platforms.

src/openxr/OpenXRApi.cpp
```diff
@@ -852,9 +893,19 @@ bool OpenXRApi::initialiseSession() {
 		return false;
 	}
 
+#ifdef ANDROID
+	XrSystemColorSpacePropertiesFB color_space_properties_fb = {
```
BastiaanOlij, line 897: I wouldn't switch this on Android but based on whether the color space extension is available. Other vendors may end up implementing this extension as well

src/openxr/OpenXRApi.cpp
BastiaanOlij: On that subject, some of these should be checked regardless and we simply check whether the extension is available or not. If it is available we make use of it, if not we should act accordingly. For instance `XR_FB_DISPLAY_REFRESH_RATE_EXTENSION_NAME` not being available isn't severe enough to fail initialising the plugin, it just means we can't query nor set the refresh rate. But as more vendors implement this extension it becomes usable in those runtimes that make it available.

src/openxr/OpenXRApi.cpp
```diff
@@ -1161,13 +1213,21 @@ bool OpenXRApi::initialiseSwapChains() {
 			.mipCount = 1,
 		};
 
+#ifdef ANDROID
+		// Enable foveation on this swapchain
```
BastiaanOlij, line 1217: Same with this, we should switch it based on the functionality being available, not on platform.

src/openxr/OpenXRApi.cpp
```diff
@@ -2502,16 +2582,22 @@ int OpenXRApi::get_external_texture_for_eye(int eye, bool *has_support) {
...
 }
 
-void OpenXRApi::process_openxr() {
-	if (!initialised) {
-		return;
-	}
-
-	XrResult result;
+const char *session_states[] = {
+		"XR_SESSION_STATE_UNKNOWN",
```
BastiaanOlij, line 2586: Hmm, wonder if we should have this at the top of our source, though I doubt we'll be using it in many places.

src/openxr/OpenXRApi.cpp
m4gr3d: Let me address this in another PR. I want to see a mechanism for required extensions, and another for optional ones which would trigger a flag based on availability.
I'll add a `TODO` for the time being.

src/openxr/OpenXRApi.cpp
m4gr3d: Let me add a todo for this one as well. It's switched on Android for now because it's one of the required extension in the Android switch above.
Once we switch to the new mechanism, then it'll be setup accordingly.

src/openxr/OpenXRApi.cpp
m4gr3d: Same here, I'll add a todo and address it in another PR.

src/openxr/OpenXRApi.cpp
BastiaanOlij: Sounds good to me

BastiaanOlij (approved):
Few things we discussed that will move forward, thanks for the help, can't wait to test how well things start running on a quest :)

PR #85, opened 2021-09-08
Move GDNative classes into their own subfolder

BastiaanOlij:
This is just some cleanup, I've moved the classes that are exposed to Godot into their own subfolder. 
The source code that remains in the `openxr` folder is now purely our implementation of the OpenXR system.

I think there is an argument to be made to rename the `OpenXRAPI` class to something else because it looks too similar to the naming we've applied to the exposed classes.

Finally I've added a placeholder set of files for `XRInterfaceOpenXR` which will replace `ARVRInterface` in Godot 4. I've left the `ARVRInterface` implementation in the root of the source folder because at this time this is not strictly speaking a class and it will be completely rewritten for Godot 4 anyway.

### DIFF
--- a/src/gdclasses/OpenXRConfig.cpp
+++ b/src/gdclasses/OpenXRConfig.cpp
@@ -1,7 +1,7 @@
 /////////////////////////////////////////////////////////////////////////////////////
 // Our OpenXR config GDNative object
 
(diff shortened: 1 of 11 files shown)


m4gr3d (approved):
Looks good!

PR #87, opened 2021-09-08
Improved extension checking logic

BastiaanOlij:
This changes the logic that checks and enabled extensions in our plugin.
It works with a `map` to which we add the extensions we want.
The entry then points to a boolean that the logic either sets or clears depending on whether the extension is available.
If this pointer is a `nullptr` we deem the extension as a requirement and it not being available will fail initialization of the plugin.
Finally we enabled all extensions that were requested and found.

The boolean flags allow further logic to actually implement the usage of the extension.

I've also moved most of the FB extensions out of the Android only sphere, I suspect a number of these will become extensions also implemented by other vendors as has happened with a number of Microsoft and Valve extensions that still bear their names (especially around refresh rates and such, really weird those aren't part of the core spec already).

### DIFF
--- a/src/ARVRInterface.cpp
+++ b/src/ARVRInterface.cpp
@@ -2,6 +2,7 @@
 // Our main ARVRInterface code for our OpenXR GDNative module
 
 #include "ARVRInterface.h"
+#include <ARVRInterface.hpp>
 #include <MainLoop.hpp>
 
 typedef struct arvr_data_struct {
@@ -22,7 +23,19 @@ godot_string godot_arvr_get_name(const void *p_data) {
 
 godot_int godot_arvr_get_capabilities(const void *p_data) {
 	godot_int ret;
-	ret = 2 + 8; // 2 = ARVR_STEREO, 8 = ARVR_EXTERNAL
+
+	// These are capabilities supported by our interface class, not necesarily by the device we're currently using.
+	// We'll need to figure out a way to query OpenXR about this.
+
+	ret = godot::ARVRInterface::ARVR_MONO; // We support mono devices (phones, etc.)
+	ret += godot::ARVRInterface::ARVR_STEREO; // We support stereo devices (HMDs)
+	// ret += godot::ARVRInterface::ARVR_QUAD; // Once we have the ability to do this, we can add it as a feature, Godot 4 most likely
+	ret += godot::ARVRInterface::ARVR_AR; // We support AR
+	// ret += godot::ARVRInterface::ARVR_VR; // We support VR, strangely missing as a flag, we'll add this in Godot 4
+#ifndef ANDROID
+	ret += godot::ARVRInterface::ARVR_EXTERNAL; // Rendering to an external device, Godot window is separate.
+#endif
+
 	return ret;
 }
 
@@ -39,11 +52,39 @@ void godot_arvr_set_anchor_detection_is_enabled(void *p_data, bool p_enable) {
 }
 
 godot_bool godot_arvr_is_stereo(const void *p_data) {
+	arvr_data_struct *arvr_data = (arvr_data_struct *)p_data;
 	godot_bool ret;
 
 	// TODO we should check our configuration and see if we are setup for stereo (hmd) or mono output (tablet)
-
-	ret = true;
+	if (arvr_data == nullptr || arvr_data->openxr_api == nullptr) {
+		ret = true;
+	} else {
+		XrViewConfigurationType config_type = arvr_data->openxr_api->get_view_configuration_type();
+		switch (config_type) {
+			case XR_VIEW_CONFIGURATION_TYPE_PRIMARY_MONO: {
+				// In Godot 4 we'll return 1
+				ret = false;
+			}; break;
+			case XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO: {
+				// In Godot 4 we'll return 2
+				ret = true;
+			}; break;
+			/* These are not (yet) supported, adding them in here for future reference
+			case XR_VIEW_CONFIGURATION_TYPE_PRIMARY_QUAD_VARJO: {
+				// In Godot 4 we'll return 4 once we add support for Quad rendering
+				return ???;
+			}; break;
+			case XR_VIEW_CONFIGURATION_TYPE_SECONDARY_MONO_FIRST_PERSON_OBSERVER_MSFT: {
+				// In Godot 4 we'll return 1, we'll need to support multiple viewports for this with alternative camera tracking.
+				return ???;
+			}; break;
+			*/
+			default: {
+				godot::Godot::print_error(godot::String("Unsupported view configuration type set: ") + godot::String::num_int64(config_type), __FUNCTION__, __FILE__, __LINE__);
+				ret = true; // we need to return something even though it really is undefined...
+			}; break;
+		}
+	}
 
 	return ret;
 }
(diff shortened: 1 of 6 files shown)


BastiaanOlij:
I also snuck in setting the view configuration type...

src/ARVRInterface.cpp
```diff
@@ -39,11 +39,25 @@ void godot_arvr_set_anchor_detection_is_enabled(void *p_data, bool p_enable) {
...
+		switch (arvr_data->openxr_api->get_view_configuration_type()) {
+			case XR_VIEW_CONFIGURATION_TYPE_PRIMARY_MONO: {
+				ret = false;
+			}; break;
+			case XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO: {
+				ret = true;
+			}; break;
+			default: {
+				ret = true;
+			}; break;
```
m4gr3d, line 58: Whenever the cases are the same, I usually combine them as it also serves as a description to the logic (e.g: default value is `XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO`).

I think we can apply the same style here.

src/openxr/OpenXRApi.cpp
```diff
@@ -910,29 +882,31 @@ bool OpenXRApi::initialiseSession() {
 		return false;
 	}
 
-#ifdef ANDROID
-	// TODO: Instead of switching on Android, use the new validation mechanism to determine when this should be configured.
+	// always define, we ignore this if it's not used.
 	XrSystemColorSpacePropertiesFB color_space_properties_fb = {
 		.type = XR_TYPE_SYSTEM_COLOR_SPACE_PROPERTIES_FB,
 	};
```
m4gr3d, line 888: Here and below, shouldn't the `XR` structs be defined under their corresponding `if` block?

src/openxr/OpenXRConfig.cpp
```diff
@@ -45,6 +51,44 @@ bool OpenXRConfig::keep_3d_linear() const {
...
+			}; break;
+			case XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO: {
+				return 1;
+			}; break;
+			default: {
+				// we don't support the others (yet)
+				return 1;
+			}; break;
+		}
+	}
```
m4gr3d, line 70: Do we have an enum that defines the configuration types? I thought they were matching the `ARVRInterface::Capabilities` enum but they're off by 1.

src/openxr/OpenXRConfig.cpp
```diff
@@ -45,6 +51,44 @@ bool OpenXRConfig::keep_3d_linear() const {
...
+		Godot::print("OpenXR object wasn't constructed.");
+	} else {
+		switch (p_config_type) {
+			case 0: {
+				openxr_api->set_view_configuration_type(XR_VIEW_CONFIGURATION_TYPE_PRIMARY_MONO);
+			}; break;
+			case 1: {
+				openxr_api->set_view_configuration_type(XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO);
+			}; break;
+			default: {
```
m4gr3d, line 85: Similar comment here about merging `default` with `case 1`.

m4gr3d (changes requested):
The overall structure looks good! I've added some feedback on some implementation details.

src/ARVRInterface.cpp
BastiaanOlij: Tricky, I like the explicitness of having every supported value as it's own case, that way it is clearly communicated what the intention is for that case and not that an option was forgotten and the default logic is in fact not correct.

This is also why I left out the two options we currently do not support.

src/openxr/OpenXRApi.cpp
BastiaanOlij: Uhm, no, because they go out of scope before they are used and that will crash the application. I've made that mistake a few times :P

It's a little ugly that as a result unused structs are defined but it is what it is.

src/openxr/OpenXRConfig.cpp
BastiaanOlij: The `ARVRInterface::Capabilities` are flags that communicate to the user what possibilities the `XRInterface` has. i.e. in this case just to communicate that stereo support is available (not enforced).

The reasons I'm using a switch here are:
- enums in Godot start with 0 so our dropdown in the UI works properly, for some reason OpenXR starts at 1
- `XR_VIEW_CONFIGURATION_TYPE_PRIMARY_QUAD_VARJO` which we currently do not support has value `1000037000`, if we are going to support that some day I'd like that to be `2`. Same deal with `XR_VIEW_CONFIGURATION_TYPE_SECONDARY_MONO_FIRST_PERSON_OBSERVER_MSFT` which if we decide to support it would likely becomes `3`.

src/openxr/OpenXRConfig.cpp
BastiaanOlij: As said, I like the explicitness here. Especially once we get more options it becomes clear what behavior is designed, but it's really nothing more then personal preference :)

src/ARVRInterface.cpp
BastiaanOlij: I've added placeholder code for the unsupported options for future reference, and added printing an error on the default code path to make clear an unsupported type was provided.

src/openxr/OpenXRConfig.cpp
BastiaanOlij: Here to I added commented out placeholder code to make it clearer what the future possibilities are and how the code will change in Godot 4 which now has the foundation for supporting multiple views. 

For Varjo we still need to implement quad rendering, for observer we'll need to add support for classifying different usages for different viewports and defining an extra camera for this purpose. Those are all future features to look at.

src/ARVRInterface.cpp
m4gr3d: Sounds good.

src/openxr/OpenXRApi.cpp
m4gr3d: I wasn't aware, thanks for the tip!

src/gdclasses/OpenXRConfig.cpp
```diff
@@ -45,6 +51,55 @@ bool OpenXRConfig::keep_3d_linear() const {
...
+			}; break;
+			*/
+			default: {
+				// we don't support the others
+				Godot::print_error(String("Unsupported configuration type set: ") + String::num_int64(config_type), __FUNCTION__, __FILE__, __LINE__);
+				return 1;
+			}; break;
+		}
+	}
+}
```
m4gr3d, line 81: Shouldn't we just return the XR view configuration types values as is? What's the advantage for defining another set of values that map to them especially since the plan is to include as much of OpenXR as possible in Godot 4.x?

m4gr3d (changes requested):

src/gdclasses/OpenXRConfig.cpp
m4gr3d: I missed this comment which seems to make the case for it: `so our dropdown in the UI works properly`.

Seems fine by me if that's the only available option.

m4gr3d (approved):

src/gdclasses/OpenXRConfig.cpp
BastiaanOlij: Yeah it's just that for a enum dropdown in the UI you define the property like so:
```
	register_property<OpenXRConfig, int>("view_configuration", ... , GODOT_PROPERTY_HINT_ENUM, "Mono,Stereo");
```
So 0 = Mono, 1 = Stereo, etc.

Also at some point we will have to think of how we'll unify some of this in Godot 4. We'll likely end up with our own enumerations in the core of Godot for this that map to whatever systems and capabilities that are in place in the runtime. While I'd like to think everyone will switch to OpenXR and we don't have to worry about any other runtimes, I think that is wishful thinking.

PR #88, opened 2021-09-12
Implement refresh rate API

BastiaanOlij:
This PR implements an API to getting/setting refresh rates on the HMD.

These functions currently only work on the Oculus runtime, only FB has implemented an extension for this, hopefully other vendors will follow suit and add this to their OpenXR implementation as well either through the same extension or a core extension. It's important that we assume this will happen some day and expose an API to the end user that is vendor agnostic.

The new feature exposes a number of functions and properties through OpenXRConfig (note that in Godot 4 I intend to move them into the XRInterfaceOpenXR class and use proper groups).
The functions added are:
`get_refresh_rate` : Returns the current refresh rate for the HMD (return 0 if none provided or not supported)
`set_refresh_rate` : Will attempt to change the refresh rate on the HMD (will be ignore if not supported or if an invalid value is provided)
`get_available_refresh_rates` : Returns an array of refresh rates supported by the HMD (empty if not supported)

The following property has been added:
`refresh_rate` : The refresh rate for the HMD, this is a runtime only property.

As part of this PR I've also moved some of the hand tracking code into separate files in the same way as I've done this for the refresh rate. Currently this is only the code for obtaining function pointers and implementing access to these functions. Turns out the OpenXR loader does not provide this functionality for extensions. 

I think more of the hand tracking and refresh rate logic can eventually be moved into these files to better segregate the logic instead of `OpenXRApi` just growing and growing but that is something for another day.

### DIFF
--- a/demo/Main.tscn
+++ b/demo/Main.tscn
@@ -43,6 +43,9 @@ directional_shadow_max_distance = 20.0
 
 [node name="FPSController" parent="." instance=ExtResource( 4 )]
 
+[node name="Configuration" parent="FPSController" index="0"]
+refresh_rate = 0.0
+
 [node name="ShadowHead" type="MeshInstance" parent="FPSController/ARVRCamera" index="0"]
 transform = Transform( 1, 0, 0, 0, -1.62921e-07, -1, 0, 1, -1.62921e-07, 0, -0.0187781, 0.103895 )
 layers = 524288
--- a/src/gdclasses/OpenXRConfig.h
+++ b/src/gdclasses/OpenXRConfig.h
@@ -30,6 +30,10 @@ public:
 	int get_form_factor() const;
 	void set_form_factor(const int p_form_factor);
 
+	double get_refresh_rate() const;
+	void set_refresh_rate(const double p_refresh_rate);
+	godot::Array get_available_refresh_rates() const;
+
 	godot::Array get_enabled_extensions() const;
 
 	String get_action_sets() const;
--- a/src/openxr/OpenXRApi.cpp
+++ b/src/openxr/OpenXRApi.cpp
@@ -15,46 +15,6 @@
 using namespace godot;
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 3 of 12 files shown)


src/openxr/OpenXRApi.cpp
```diff
@@ -833,29 +793,27 @@ bool OpenXRApi::initialiseInstance() {
 	return true;
 }
 
-bool OpenXRApi::initialiseExtensions() {
+bool OpenXRApi::initialise_extensions() {
 	XrResult result;
```
BastiaanOlij, line 797: Note, I'm slowly renaming things according to our naming convention discussion. I'll probably do a PR soon that goes through most of the current code base but for now I'm only concentrating on things impacted by this PR.

demo/Main.tscn
```diff
@@ -43,6 +43,9 @@ directional_shadow_max_distance = 20.0
 
 [node name="FPSController" parent="." instance=ExtResource( 4 )]
 
+[node name="Configuration" parent="FPSController" index="0"]
+refresh_rate = 0.0
+
```
BastiaanOlij, line 48: This looks like a bug in GDNative, this property is set to not be available in the editor.. alas can't do much harm.

src/gdclasses/OpenXRConfig.h
```diff
@@ -30,6 +30,10 @@ class OpenXRConfig : public Node {
 	int get_form_factor() const;
 	void set_form_factor(const int p_form_factor);
 
+	double get_refresh_rate() const;
+	void set_refresh_rate(const double p_refresh_rate);
```
BastiaanOlij, line 34: Just in case anyone asks, I'm using doubles here even though floats are used in the API. This is because with the new GD extensions everything is converted to doubles, might as well start using that as a base.

m4gr3d (approved):

BastiaanOlij:
Just in case anyone was wondering, this worked a treat on my quest 1 :)

PR #89, opened 2021-09-12
Improve plugin lifecycle callbacks

m4gr3d:
Update the architecture to initialize and integrate OpenXR extensions within Godot's OpenXR lifecycle.


src/openxr/OpenXRApi.cpp
```diff
@@ -1534,21 +1505,26 @@ void OpenXRApi::uninitialize() {
 	if (session != XR_NULL_HANDLE) {
 		xrDestroySession(session);
 		session = XR_NULL_HANDLE;
+
+		for (XRExtensionWrapper * wrapper : registered_extension_wrappers) {
+			wrapper->on_session_destroyed();
+		}
```
BastiaanOlij, line 1511: I wonder if we need to call the extension code before actually destroying the session, we may be cleaning things up that still need to have the session available. Not sure to be honest...

src/openxr/OpenXRApi.cpp
```diff
@@ -1534,21 +1505,26 @@ void OpenXRApi::uninitialize() {
...
 	}
 	if (instance != XR_NULL_HANDLE) {
 		xrDestroyInstance(instance);
 		instance = XR_NULL_HANDLE;
+
+		for (XRExtensionWrapper * wrapper : registered_extension_wrappers) {
```
BastiaanOlij, line 1517: Same here, should we be calling before destroying the instance as we may need the instance to clean up...

src/openxr/extensions/xr_ext_hand_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+    result = xrGetSystemProperties(openxr_api->get_instance(), openxr_api->get_system_id(), &systemProperties);
+    if (!openxr_api->xr_result(result, "Failed to obtain hand tracking information")) {
+        return false;
+    }
+
+    if (!handTrackingSystemProperties.supportsHandTracking) {
+        // The system does not support hand tracking
+        printf("Hand tracking is not supported\n");
+        return false;
```
BastiaanOlij, line 150: Something I never cleaned up but this should be using `Godot::print`, nitpicking :)

src/openxr/extensions/xr_fb_display_refresh_rate_extension_wrapper.cpp
```diff
@@ -0,0 +1,28 @@
...
+    OpenXRApi::openxr_release_api();
+}
+
+std::map<const char *, bool *> XRFbDisplayRefreshRateExtensionWrapper::get_request_extensions() {
+    return request_extensions;
+}
+
+void XRFbDisplayRefreshRateExtensionWrapper::cleanup() {
+
+}
```
BastiaanOlij, line 28: Was the rest of the implementation missing here for a specific reason or did you not get around to it? Happy to do this once merged.

src/openxr/OpenXRApi.h
```diff
@@ -275,6 +250,8 @@ class OpenXRApi {
 	OpenXRApi();
 	~OpenXRApi();
 
+	void register_extension_wrapper(XRExtensionWrapper* wrapper);
+
```
BastiaanOlij, line 254: Where is this called to actually register our two wrappers? I'm probably overlooking the code but I don't see it...

m4gr3d: It's invoked in `ARVRInterface#godot_arvr_initialize()` for now. This can be moved around if the initialize call sequence for Godot 4.x will be significantly different.

src/openxr/OpenXRApi.cpp
m4gr3d: Since the opposite callback is invoked after the session is initialized, it makes sense to have this one invoked prior to the session being destroyed. I'll swap the order.

src/openxr/extensions/xr_ext_hand_tracking_extension_wrapper.cpp
m4gr3d: Done!

src/openxr/OpenXRApi.h
BastiaanOlij: Ok cool, I guess that wasn't checked in yet or I missed it :)

src/ARVRInterface.cpp
```diff
@@ -113,6 +115,10 @@ godot_bool godot_arvr_initialize(void *p_data) {
 
 	// We (already) have our API instance? cool!
 	if (arvr_data->openxr_api != nullptr) {
+		// Register our extensions
+		arvr_data->openxr_api->register_extension_wrapper<XRFbDisplayRefreshRateExtensionWrapper>();
```
BastiaanOlij, line 119: While I hope nobody does this in practice, in theory someone could uninitialise the interface and re-initialise it in which case we'll double register the extension wrappers. We may want to either make sure the logic ignores wrappers that have already been registered or find another way to only execute this code on the first initialize call.

BastiaanOlij (approved):
Much cleaner and a good base to build on top of for other extensions, thanks Fredia! Merge whenever you're happy

src/ARVRInterface.cpp
m4gr3d: I switched from a `std::vector` to a `std::set` to prevent extension wrappers from being registered more than once.

PR #90, opened 2021-09-13
Implement FB color space extension

BastiaanOlij:
Redid the color space extension implementation using the new extension system.


BastiaanOlij:
This is nearly done, getting a crash on `get_system_properties` for the new extension, it doesn't seem to like the way we push the pointers around even though our structure should remain in memory seeing we never destroy our singleton once created...

m4gr3d (approved):

PR #92, opened 2021-09-14
Port to godot 4.0 - first steps

BastiaanOlij:
Taking our first steps to porting this plugin to Godot 4 based on the work done in the godot-xr-reference implementation and the godot_openvr plugin. 

Currently relies on:
- https://github.com/godotengine/godot/pull/51179

To try out the work so far:
- build Godot with the above mentioned PR
- run `godot ----dump-extension-api`
- copy the generated `extension_api.json` into `thirdparty/godot-cpp/godot-headers`
- build `thirdparty/godot-cpp` using `release=debug`
- build the plugin using `release=debug`


m4gr3d:
@BastiaanOlij Looking at some of the changes, it looks like we're gonna need a separate demo for Godot 4.x since that won't be compatible with Godot 3.x.
Also similar for `godot-cpp`; we're going to need a Godot 3.x `godot-cpp` dependency, and a Godot 4.x `godot-cpp` dependency.

BastiaanOlij:
We'll manage it with branches. The source is going to drastically change between Godot 3 and Godot 4, there is no way we can keep them in one branch.

So plugin versions 1.x.y is for Godot 3, and plugin versions 2.x.y for Godot 4.

Right now master is still 1.x.y and we have the 4.0-dev branch for the new implementation, but eventually we'll maintain 1.x-dev branch for Godot 3 and master becomes Godot 4.

BastiaanOlij:
While this is not working and far from finished, thinking about merging the work so far into the `4.0-dev` branch so @ChristophHaag and I can work in parallel and issue PRs. 

There willl be days where we'll need to rebase 4.0-dev on master which will be a PITA but it already is a PITA as it stands right now so I think the benefits outweigh the negatives. 

@m4gr3d what do you think?

m4gr3d:
@BastiaanOlij I agree with merging into the dev branch. It should improve on the current state of development and enable additional eyes on the code.

BastiaanOlij:
I've got a few more things to push in later today and than I will merge this. Its far from working but it means people can start fixing things. 
I'm focusing on rewriting the action map stuff atm so thats all half done but mostly disabled. I'm running in a bunch of roadblocks because extensions is just not finished enough, but groud and bruvzg are helping out now so we should be able to push this forward.

PR #93, opened 2021-09-14
Setup Android api layer

m4gr3d:
- The first commit puts the foundation in place to add the Android layer version of the Godot OpenXR api. It's mostly empty for now asides from Godot's plugin setup utilities.
- The second commit adds an Android module which can be used to quickly build and run the demo project from within Android Studio. The demo project itself is not modified by the process. This will help facilitate and speed up development/debugging of the plugin.


m4gr3d:
@BastiaanOlij Is it possible to omit a directory from `scons`?
The `src/jni` directory added in this PR needs to be omitted since it's only valid for Android builds.

BastiaanOlij:
> 
> 
> @BastiaanOlij Is it possible to omit a directory from `scons`?
> The `src/jni` directory added in this PR needs to be omitted since it's only valid for Android builds.

Already mentioned this in chat but for completeness sake. For scons you just need to write out [these three lines](https://github.com/GodotVR/godot_openxr/blob/master/SConstruct#L155) in more detail instead of relying on wildchars:
```
sources += Glob('src/*.cpp')
sources += Glob('src/*/*.cpp')
sources += Glob('src/*/*/*.cpp')
```

And make similar changes for cmake:
https://github.com/GodotVR/godot_openxr/blob/master/android/CMakeLists.txt#L104

BastiaanOlij:
Also can you add some info to the readme about debugging from Android Studio. I always forget the details about this :)

BastiaanOlij (approved):
Very cool!

PR #95, opened 2021-09-20
Misc fixes

m4gr3d:
- Add utility header file
- Fix the composition layer flags to allow alpha blend
- Introduce the `XRCompositionLayerProvider` to allow custom layer providers.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,8 +9,8 @@ android {
     defaultConfig {
         minSdkVersion versions.minSdk
         targetSdkVersion versions.targetSdk
-        versionCode 10004
-        versionName '1.0.4'
+        versionCode 10005
+        versionName '1.0.5'
         setProperty("archivesBaseName", "godotopenxr.${versionName}")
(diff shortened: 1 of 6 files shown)


BastiaanOlij (approved):
More then happy with the layer implementation, I'm sure it will grow as we implement other uses for the layers. Ben's been wanting to port his functionality from the OpenVR plugin so we'll see how that ties in. My only remark here is always enabling the ALPHA bit. We may need to switch that at some point. There might be a small performance impact here as the GPU suddenly needs to read fragments and blend. 

Will have to see the macros in action but I can see the value in simplifying some of the code there. Its a bit of a shame the loaders don't just get all the extensions function pointers for us..

Lastly I think the `get_viewport` should be renamed to `get_main_viewport`. Won't know for sure until I see the intended use but it's important to know that on desktop VR there is no guarantee the main viewport is used as the VR render target, and when layer functionality evolves we're likely to have multiple viewports involved.

PR #97, opened 2021-09-27
Add more extension wrappers

m4gr3d:
- Add wrapper implementation for the `XR_EXT_performance_settings` extension
- Add wrapper implementation for the `XR_FB_foveation` extension.
- Store the system name and vendor id.
- Add api to customize the render target size.


src/openxr/OpenXRApi.cpp
```diff
@@ -2546,7 +2568,13 @@ bool OpenXRApi::poll_events() {
 				// TODO: do something
 			} break;
 			default:
-				Godot::print_warning(String("OpenXR Unhandled event type ") + String::num_int64(runtimeEvent.type), __FUNCTION__, __FILE__, __LINE__);
+				bool handled = false;
+				for (XRExtensionWrapper *wrapper : registered_extension_wrappers) {
```
BastiaanOlij, line 2572: I would place this above our own switch so the extensions can also react to events that we're already reacting too. 

So it becomes:
```
bool handled = false;
for (XRExtensionWrapper *wrapper : registered_extension_wrappers) {
	handled |= wrapper->on_event_polled(runtimeEvent);
}
switch (runtimeEvent.type) {
	...
	default: {
		if (!handled) {
			Godot::print_warning(String("OpenXR Unhandled event type ") + String::num_int64(runtimeEvent.type), __FUNCTION__, __FILE__, __LINE__);
		}
 	} break;
 }
```
All though admittedly it may be a little double seeing we call into the wrapper for some of the events already.

Also small nitpick but it's up to you if you want to follow suit, I always add a code block in each case of a switch including (so `default: { ... } break;`). It's a style of programming where you assume people who come after you may not understand your code fully and make a change earlier on in code that break because everything is in the same scope. It also enforces a visual consistency making anything that breaks the pattern stand out. 
I do the same with `if` statements, you'll rarely see me do one lines instead always writing out the if statement block fully as `if (condition) { code }`

Again it's really something I just adopted a long time ago so I don't expect anyone to follow suit, but seeing it breaks the pattern if someone doesn't it always stands out to me :)

src/openxr/OpenXRApi.h
```diff
@@ -285,6 +284,13 @@ class OpenXRApi {
 	XrSystemId get_system_id() { return systemId; }
 	XrSpace get_play_space() { return play_space; }
 	XrFrameState get_frame_state() { return frameState; }
+	String get_system_name() const { return system_name; }
+	uint32_t get_vendor_id() const { return vendor_id; }
+
+	bool set_render_target_size_multiplier(float multiplier);
```
BastiaanOlij, line 290: I think we should also add a `get_render_target_size_multiplier` and expose this as a property on our `OpenXRConfig` object, it's been a much requested feature.

src/openxr/OpenXRApi.cpp
```diff
@@ -953,6 +960,21 @@ bool OpenXRApi::initialiseSession() {
 	return true;
 }
 
+bool OpenXRApi::set_render_target_size_multiplier(float multiplier) {
+	if (is_initialised()) {
+		Godot::print_error("Setting the render target size multiplier is only allowed prior to initialization.", __FUNCTION__, __FILE__, __LINE__);
```
BastiaanOlij, line 965: Probably not for now but I think we need to look into what it takes to recreate the swapchain if anything changes that requires a differently sized swap chain. 

SteamVR for instance has a mode where it will periodically tweak the size of the render target depending on system load, so if it finds out you're consistently not reaching steady FPS, it will request a differently sized render target.

OpenXR doesn't have an event to notify us of this yet and we could only implement this by polling but it's something to start thinking about. Again not for this PR but for later.

Also there is obviously a clash between XR runtimes that already support a multiplier as a feature and us now having our own that is applied on top of it but c'est la vie.

BastiaanOlij:
On the update swapchain state extension, I could be mixing things up but doesn't this introduce a new event that informs us of state changes related to the swapchain? It looks like we enable the extension but don't actually use it. I'm assuming this comes later?

BastiaanOlij (approved):
Looks good, my only remarks are nitpicking or just wondering how things worked. Feel free to merge, we can always take things on board later.

The only one you may consider right away if it makes sense, is the change I suggested around handling events and sending all events to the active wrapper classes.

src/openxr/OpenXRApi.cpp
m4gr3d: Agreed.
Similar to that, we're also assuming that all the `XrViewConfigurationView` properties will have matching render target dimensions and swapchain sample count. This may not always be the case (e.g: devices with more than 2 views), so here also this is something we may need to address in the future.

src/openxr/OpenXRApi.cpp
m4gr3d: That's a good call actually. I've myself ran into this issue several times.

src/openxr/OpenXRApi.h
m4gr3d: Done!
I'm only adding the core logic in this PR. I'll add the matching gdclasses properties and methods in a separate PR.

m4gr3d:
> On foveated rendering, I've always wondered about the technical details, normally it entails setup on the frame buffer to actually make this work so Oculus must be doing some sort of magic to embed it in the texture itself. While we've worked around it to some degree in the GLES2 renderer this approach assumes you use a single render buffer and just render into that, this makes it completely impossible to implement certain effects.
> 
> We need to understand more of this (or I do, you may have read up on this way further then I have) because with the Vulkan renderer will be rendering into our own HDR buffers and then rendering into the final buffers as we apply tonemapping and post effects. As we're using subpasses now we don't incur the performance penalty we did with the GLES3 renderer (well depending on which post processes the user enables, some can't be done with subpasses).
> Interestingly enough Oculus had a blog post a little while ago where they recommended the same approach.
> 
> From what I've read so far OpenXR ties into Vulkan VRS to do foveated rendering where we request the VRS map and supply that when setting up our framebuffer. I have yet to dive into it fully but thats the plan so far. I wonder how that works together with Oculus' own extensions.

I plan to follow up with that topic at some point in the future. Currently their OpenXR foveation implementation doesn't seem to be fully working (both in Godot and in their own sample demo), so I'm following up with that at the moment to figure why that's the case.
Once that issue is resolved, we can approach them with that next topic.

m4gr3d:
> On the update swapchain state extension, I could be mixing things up but doesn't this introduce a new event that informs us of state changes related to the swapchain? It looks like we enable the extension but don't actually use it. I'm assuming this comes later?

It gives us the ability to query the current state, but no events yet to be notified of changes to it: https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_FB_swapchain_update_state

For our logic, the extension is required to enabled foveation on the swapchain.

PR #101, opened 2021-09-30
XR_EXTX_Overlay support

ChristophHaag:
SteamVR does not implement this extension, only tested on Monado.

Open questions:
* public variables on the extension wrapper okay?
* should main_session_visible be a NO_EDITOR property instead of calling get_main_session_visible every _process()?
* `env.background_color.a = 0.0` doesn't actually make the background transparent


ChristophHaag:
New problem: A call to `get_tree().get_root().set_transparent_background(true)`  or `get_tree().get_root().set_transparent_background(false)` breaks VR rendering, produces weird flickering of wrong/old textures.

BastiaanOlij:
> 
> 
> New problem: A call to `get_tree().get_root().set_transparent_background(true)` or `get_tree().get_root().set_transparent_background(false)` breaks VR rendering, produces weird flickering of wrong/old textures.

The problem here is that this triggers resizing the viewport back to its original dimensions, for some reason this breaks when used in combination with render targets supplied by the VR plugin, I've got it on my list to look into but haven't gotten to it yet.

A viable workaround is to not use the main viewport but a separate viewport for VR rendering, only works for desktop VR though. You can have a look at the demo on the OpenVR plugin which uses this approach.

PR #102, opened 2021-09-30
Add support for touch detection on A/X and B/Y buttons.

m4gr3d:


BastiaanOlij:
You are missing the changes to the jsons to add the needed actions to make this work.

That said, what is the use case for detecting touch on these buttons? If there is a valid one unfortunately we need to add this way but otherwise its a lot of added clutter. We'd need to start adding touch for trigger, grip, etc. as well.

It was always my understanding that these sensors were solely used to animate the hand model to give visual feedback. That in OpenXR is now solved by returning the animated hand skeleton through hand tracking (though we're having issues with that on Quest atm).

The touch sensors for the trackpad are needed because you only want the logic using trackpad to react visually when the user is engaging with the trackpad and ignore the trackpad when the user is not. 

Wish we could renumber the buttons but that would be a major breaking change at this point we can ill afford.

BastiaanOlij (approved):
As per my remark, I would like to know the use case as it feels like clutter, but if there is a valid use case then the work is sounds and this can be merged.

m4gr3d:
> That said, what is the use case for detecting touch on these buttons? If there is a valid one unfortunately we need to add this way but otherwise its a lot of added clutter. We'd need to start adding touch for trigger, grip, etc. as well.
>
> It was always my understanding that these sensors were solely used to animate the hand model to give visual feedback. That in OpenXR is now solved by returning the animated hand skeleton through hand tracking (though we're having issues with that on Quest atm).

As discussed offline, the use case is for animating the hand model given that, at least on Quest, visual hand tracking and controller tracking are mutually exclusive and the hand model is only automatically animated when visual hand tracking is active.
For animating the hand model when controller tracking is active, the use of the added `touch` sensors is required.

PR #114, opened 2021-10-28
Add gdnlib config for armeabi-v7a

m4gr3d:


BastiaanOlij:
Should we also add building this to the CI?

m4gr3d:
> Should we also add building this to the CI?

If it's straightforward, then yes, let's add it to the CI.

PR #117, opened 2021-11-03
Move all thirdparty libraries into a thirdparty folder

BastiaanOlij:
Moving all 3rd party libraries into a separate folder. I need to add a couple more for Vulkan, this way it's easier to keep track of them and update them.

Also I'm thinking about dropping the version number from the folder names so there is less chance we mess up paths when updating to a newer version, what do you guys reckon?


m4gr3d (approved):

BastiaanOlij:
Ok cleaned things up some more, godot-cpp is now in the thirdparty folder, removed the version numbers. Updated the readme to reflect the version numbers. All set to go

PR #124, opened 2021-11-17
Fix hand tracking on the Oculus Quest

### LINKED ISSUE #82: Skeleton/finger tracking on Quest
We've disabled hand tracking on Android because this currently crashes on the Quest.

This is purely used for skeletal animation of the hand either by inferring the positioning based on controller sensors, or for full hand tracking as supported on the quest.

m4gr3d:
The issue was caused by an improperly initialize `next` pointer.

Fix #82

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@ compile_commands.json
 
 # Ignore Gradle build output directory
 android/build
-android/.cxx
+.cxx
 
 # Ignore intellij configuration files and directories
 .idea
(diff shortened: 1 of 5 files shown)


BastiaanOlij (approved):
Oops, missed the next there, with MSVC zeroing out the struct by default it wasn't a problem.. 

Using the struct initialisers to make sure any missed fields are always properly zeroed out seems like the better solution so I'm all for this fix!

PR #125, opened 2021-11-17
Add support for Oculus passthrough OpenXR extensions

m4gr3d:


m4gr3d:
@BastiaanOlij The PR has a dependency on OpenXR v1.0.19 or higher in order to properly build (hence the draft state).

I'm not familiar with the process to update the `openxr_loader` directory, so can you perform the update and add the instructions to the `README` so I can perform it in the future.

android/src/main/AndroidManifest.xml
```diff
@@ -13,6 +13,13 @@
         android:version="1"
         tools:node="replace"/>
 
+    <!--
+    Passthrough feature flag.
+    `required` should be `true` when passthrough is required for app functionality, `false` otherwise.
+    -->
+    <uses-feature android:name="com.oculus.feature.PASSTHROUGH"
+        android:required="false" />
```
BastiaanOlij, line 21: Does this mean this feature only works when use `AAR` or can we add this to Godot itself?

src/openxr/extensions/xr_fb_passthrough_extension_wrapper.cpp
```diff
@@ -0,0 +1,216 @@
...
+	// Resume the passthrough layer
+	Godot::print("Starting passthrough layer...");
+	result = xrPassthroughLayerResumeFB(passthrough_layer);
+	if (!openxr_api->xr_result(result, "Failed to start the passthrough layer")) {
+		stop_passthrough();
+		return false;
+	}
+
+	// Enable the viewport transparent background
+	Viewport *viewport = get_main_viewport();
```
BastiaanOlij, line 85: In due time we need to improve on this to detect which viewport is actually used. Once desktop headsets start supporting passthrough as well, which they undoubtedly will, it is not guaranteed the main viewport is the one that is used.

BastiaanOlij:
> 
> 
> @BastiaanOlij The PR has a dependency on OpenXR v1.0.19 or higher in order to properly build (hence the draft state).
> 
> I'm not familiar with the process to update the `openxr_loader` directory, so can you perform the update and add the instructions to the `README` so I can perform it in the future.

I'll do this tomorrow but in a nutshell, just replace the version in third party with the latest one I'm happy to go to v1.0.20 btw seeing its out.

BastiaanOlij (approved):
This looks really good, I need to find a moment to test it but very exciting!

android/src/main/AndroidManifest.xml
m4gr3d: We can this to Godot itself.

Most of the Godot mobile VR options have been for the VrApi plugin, so we need to figure out how to introduce configuration options for the OpenXR plugin in a way that doesn't confuse users (e.g: Do we deprecate the VrApi options in the next 3.x release?).

In addition, since the OpenXR plugin is multi-platform, we need to figure out a way to inform the user when a selected configuration is not available on the target platform.

BastiaanOlij: I think we need to keep supporting VrAPI until people have switched over but indeed, we need to figure out a nicer way to switch the different options in/out of the config.

As for what is and isn't supported in OpenXR itself, that one is harder especially for people who are making cross platform games. What is available on the machine they are working on, and what is available on the different target platforms they are deploying on differs.

In my eyes this is a combo between making sure features that aren't supported have good default behavior and exposing what is supported through the configuration object. Right now it already has the supported features string that we build as we enable extensions, but I think some of those need to be simplified to simple booleans we can query. So a "supports_handtracking" bool and a "supports_passthrough" bool, so the person building a game can react accordingly.

BastiaanOlij:
Don't forget to update the changes.md file @m4gr3d

m4gr3d:
Good call, i forgot to do so in the past.

BastiaanOlij:
> 
> 
> Good call, i forgot to do so in the past.

I always forget :(

PR #131, opened 2021-11-22
Updating initialisation process

BastiaanOlij:
When this plugin was originally implemented it would often be used alongside of other plugins, for instance using OpenXR for desktop and VrApi for Quest support.

As such the initialisation didn't happen automatically when you added the premade scene to your project. The idea here was that you would only call the initialisation function if you ended up using OpenXR but this does mean things do not work "out of the box".

As OpenXR is now at a stage where it can be used to deploy to all platforms it makes sense to automatically initialise it. To not remove existing logic I've added an export variable you can set in the UI which is set by default. There is an argument to do away with this completely and just initialise on `_ready`, worth a discussion here.

image

I've also added signals to communicate back to the main project whether initialisation succeeded or not so the game can do further initialisation that is required after OpenXR is finished. This also allows you to exit the application on failure.

### DIFF
--- a/demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
+++ b/demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
@@ -1,26 +1,45 @@
 extends ARVROrigin
 
+signal initialised
+signal failed_initialisation
+
+export var auto_initialise = true
+export var start_passthrough = false
 export (NodePath) var viewport = null
 
 var interface : ARVRInterface
 
+func _ready():
+	if auto_initialise:
+		initialise()
+
 func initialise() -> bool:
-	var interface = ARVRServer.find_interface("OpenXR")
+	if interface:
+		# we are already initialised
+		return true
+
+	interface = ARVRServer.find_interface("OpenXR")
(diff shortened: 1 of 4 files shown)


demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
```diff
@@ -1,27 +1,46 @@
...
 		var vp : Viewport = null
 		if viewport:
 			vp = get_node(viewport)
 
 		if !vp:
 			vp = get_viewport()
 
+		# Start passthrough?
+		if start_passthrough:
+			# make sure our viewports background is transparent
```
BastiaanOlij, line 35: @m4gr3d I've also added an export variable to enable/disable pass through. I don't think it's a good idea to suggest to the user to make changes in the files that are in the plugin. This means that whenever they update the plugin with a newer version, they need to make the changes again. With export variables the decision is recorded in their own project and it becomes easier to upgrade to newer versions.

Seeing the transparent background is a little double here but as we know the correct viewport here it seemed like a smart thing to do.

m4gr3d (approved):

PR #133, opened 2021-12-10
Add support for retrieving our playspace bounds

BastiaanOlij:
This PR adds the `get_play_space` method to `OpenXRConfig` which returns 4 points that denote a plane on the floor that fits within the players play area boundaries. 

By default this is an axis aligned rectangle centered on the origin point placed on the floor however if at any time `CenterHMD` is called it is possible that this rectangle will be rotated and moved so that it is placed at the real origin, not the adjusted origin. 

Note that it is up to the user to call `get_play_space` after `CenterHMD` is called. 
Also note that `get_play_space` will only return an usable result only after OpenXR has ran through the full initialisation process. It is best to react on the `openxr_visible_state` signal sent by the ARVRServer. 

While normally not used for visualization but to drive game logic, the demo contains an example of using `get_play_space` to visualize this play space. If `RESET_FULL_ROTATION` is used (very VERY uncommon, I doubt anyone still uses this option) with CenterHMD the play space will be slanted in such a way that this visualization will fail, it assumes the rectangle is parallel with the floor.


demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
```diff
@@ -3,12 +3,22 @@ extends ARVROrigin
 signal initialised
 signal failed_initialisation
 
+# Also add signals so we can have other parts of the application react to this.
+signal session_begun
```
BastiaanOlij, line 7: @m4gr3d I'm chaining these signals so we can easily hook them up in the UI so game code can react on them.

demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
```diff
@@ -59,7 +69,8 @@ func initialise() -> bool:
 
 func _get_xr_viewport() -> Viewport:
 	if viewport:
-		return get_node(viewport)
+		var vp : Viewport = get_node(viewport)
```
BastiaanOlij, line 72: This is just to make Godots type checking happy. This ensures the `Node` returned is properly cast to `Viewport` and then return. Without this Godot will warn about the missing cast.

src/gdclasses/OpenXRConfig.cpp
```diff
@@ -306,3 +309,37 @@ void OpenXRConfig::stop_passthrough() {
 		passthrough_wrapper->stop_passthrough();
 	}
 }
+
+godot::Array OpenXRConfig::get_play_space() {
+	ARVRServer *server = ARVRServer::get_singleton();
+	Array arr;
+	Vector3 sides[4] = {
+		Vector3(-0.5f, 0.0f, -0.5f),
```
BastiaanOlij, line 317: Ok this is where things do get weird. When reading the documentation ( https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#xrGetReferenceSpaceBoundsRect ) it seems like we're going to retrieve extents, it's really weird because on one hand they talk about extents from the origin point, and then they talk about width and height. 

For SteamVR this seems to definitely be returning width and height. But we could end up having a bounding rectangle that is half the required size if others runtimes interpret the specs differently, I definitely started off thinking these were extents.

BastiaanOlij: Ok, I think this is a mixup between the words `extent`, which means area and `extend` which means something that is increased in size I guess. And I wonder if that is a misunderstand in Godot because we have a few fields called extent that are indeed half sizes.

Reading up in a dictionary, extent covers the full area, so the usage of total width and height seems correct

BastiaanOlij:
I really really dislike not being allowed to indent an empty line, grmbl... stupid file format checks failing on gdscript files... :)

src/gdclasses/OpenXRConfig.cpp
```diff
@@ -306,3 +310,37 @@ void OpenXRConfig::stop_passthrough() {
...
+		Size2 extends = openxr_api->get_play_space_bounds();
+		if (extends.x != 0.0 && extends.y != 0.0) {
+			Transform reference_frame = server->get_reference_frame();
+
+			for (int i = 0; i < 4; i++) {
+				Vector3 coord = sides[i];
+
+				// scale it up
+				coord.x *= extends.x;
+				coord.z *= extends.y;
```
m4gr3d, line 334: Here, better to use `extends.width` and `extends.height` to avoid any confusion when reading the code (e.g: wondering why a seemingly y-axis value is assigned to `z`).

src/openxr/OpenXRApi.cpp
```diff
@@ -2021,6 +2021,25 @@ void OpenXRApi::set_form_factor(const XrFormFactor p_form_factor) {
...
+		XrExtent2Df extends;
+
+		XrResult result = xrGetReferenceSpaceBoundsRect(session, play_space_type, &extends);
+		if (!xr_result(result, "Couldn't obtain play space bounds!")) {
+			return ret;
+		}
+		// Godot::print("OpenXR obtained extends {0} {1}.", extends.width, extends.height);
+
+		ret.x = extends.width;
+		ret.y = extends.height;
```
m4gr3d, line 2037: Here also, let's use `ret.width` and `ret.height` instead.

m4gr3d (changes requested):
Looks good, just a few minor changes and it can be merged!

src/gdclasses/OpenXRConfig.cpp
BastiaanOlij: Owh I had forgotten there is a union on Vector2 (which Size2 is coming from) exposing x and y as width and height, yeah makes sense!

m4gr3d (approved):

PR #140, opened 2021-12-21
Automatically start the passthrough layer at construction

m4gr3d:


BastiaanOlij:
Is this safe? Wouldn't this have a performance penalty if you're not using passthrough?

m4gr3d:
Yeah, the passthrough layer is only created when `start_passthrough` is invoked. 

In addition if passthrough is not enabled in the export config, the manifest permission is removed which causes three passthrough feature to be disabled.

Issue #139, opened 2021-12-21
Right eye flickers when `Viewport#transparent_bg` is enabled after `XR_SESSION_STATE_READY` state


m4gr3d:
#### Repro steps:

- In [first_person_controller_vr.gd](https://github.com/GodotVR/godot_openxr/blob/master/demo/addons/godot-openxr/scenes/first_person_controller_vr.gd), add `_get_xr_viewport().transparent_bg = true` either to [`_on_openxr_session_begun`](https://github.com/GodotVR/godot_openxr/blob/780fbbbc27d2a2a9560ad5fd3415338ac1885b03/demo/addons/godot-openxr/scenes/first_person_controller_vr.gd#L92), [`_on_openxr_visible_state`](https://github.com/GodotVR/godot_openxr/blob/780fbbbc27d2a2a9560ad5fd3415338ac1885b03/demo/addons/godot-openxr/scenes/first_person_controller_vr.gd#L104) or [`_on_openxr_focused_state`](https://github.com/GodotVR/godot_openxr/blob/780fbbbc27d2a2a9560ad5fd3415338ac1885b03/demo/addons/godot-openxr/scenes/first_person_controller_vr.gd#L100).
- When the demo is run on a Quest device, the right eye flickers continuously.
- This doesn't happen if `transparent_bg` is set in [`initialise`]().

m4gr3d:
@BastiaanOlij I have only been able to test on the Quest but this could also be an issue on other platforms.

BastiaanOlij:
Strange, this is caused by Godot forgetting that we're using external textures and recreating the buffers on the Godot side. It's the same issue we had when you resized the preview window on desktop. That was fixed but possibly I only applied the fix for resizing, not when changing transparency. I'm assuming this was tested on the latest 3.x ?

ChristophHaag:
Also an issue for #101

BastiaanOlij:
Found the issue, I think something changed in the cleanup code of textures at some point. As we create a wrapper Texture object around our OpenGL texture supplied by the plugin, on changing the transparency setting this texture object is destroyed (as we recreate a new render target with proper settings for transparency) and our current right eye texture along side it. Hence the flicker as one texture in the texture chain is now missing.

This fixes it: https://github.com/godotengine/godot/pull/56148

BastiaanOlij:
@m4gr3d let me know if the fix I did fixes this for you as well, if so we can close this:)

m4gr3d:
[...]

PR #141, opened 2021-12-24
Make passthrough a toggle on first person controller

BastiaanOlij:
Right now the properly on the first person controller script just auto starts passthrough.

This change makes it a proper toggle with which to turn on/off passthrough.

@m4gr3d is this the correct way to implement this?

### DIFF
--- a/demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
+++ b/demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
@@ -1,3 +1,4 @@
+tool
 extends ARVROrigin
 
 signal initialised
@@ -10,13 +11,35 @@ signal focused_state
 signal visible_state
 signal pose_recentered
 
-export var auto_initialise = true
-export var start_passthrough = false
-export (NodePath) var viewport = null
+export var auto_initialise = true setget set_auto_initialise
+export var enable_passthrough = false setget set_enable_passthrough
+export (NodePath) var viewport setget set_viewport
 export var near_z = 0.1
 export var far_z = 1000.0
 
 var interface : ARVRInterface
+var enabled_extensions : Array
+
+func set_auto_initialise(p_new_value):
+	auto_initialise = p_new_value
+	update_configuration_warning()
+
+func set_enable_passthrough(p_new_value):
+	enable_passthrough = p_new_value
+	update_configuration_warning()
+
+	# Only actually start our passthrough if our interface has been instanced
+	# if not this will be delayed until initialise is successfully called.
+	if interface:
+		if enable_passthrough:
+			# unset enable_passthrough if we can't start it.
+			enable_passthrough = _start_passthrough()
+		else:
+			_stop_passthrough()
+
+func set_viewport(p_new_value):
+	viewport = p_new_value
+	update_configuration_warning()
 
 func get_interface() -> ARVRInterface:
 	return interface
@@ -25,10 +48,14 @@ func _ready():
 	$ARVRCamera.near = near_z
 	$ARVRCamera.far = far_z
 
-	if auto_initialise:
+	if auto_initialise && !Engine.editor_hint:
 		initialise()
 
 func initialise() -> bool:
+	if Engine.editor_hint:
+		print("Can't initialise while in the editor")
+		return false
+
 	if interface:
 		# we are already initialised
 		return true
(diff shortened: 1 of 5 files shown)


demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
```diff
@@ -35,8 +48,12 @@ func initialise() -> bool:
 		# Find the viewport we're using to render our XR output
 		var vp : Viewport = _get_xr_viewport()
 
+		# Obtain enabled extensions
+		enabled_extensions = $Configuration.get_enabled_extensions()
+
 		# Start passthrough?
-		_start_passthrough()
+		if enable_passthrough:
+			_start_passthrough()
```
m4gr3d, line 56: Here you should follow the same pattern as in `set_passthrough(...)` and update `enable_passthrough` with the result from `_start_passthrough`.

demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
```diff
@@ -74,13 +93,21 @@ func _get_xr_viewport() -> Viewport:
 	else:
 		return get_viewport()
 
-func _start_passthrough():
-	if start_passthrough:
-		# make sure our viewports background is transparent
-		_get_xr_viewport().transparent_bg = true
+func is_passthrough_supported() -> bool:
+	return enabled_extensions.find("XR_FB_passthrough") >= 0
```
m4gr3d, line 97: Is this method being used?

demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
```diff
@@ -74,13 +93,21 @@ func _get_xr_viewport() -> Viewport:
...
+
+func _start_passthrough() -> bool:
+	# make sure our viewports background is transparent
+	_get_xr_viewport().transparent_bg = true
```
m4gr3d, line 101: Is there a way to detect the version of Godot being used and print a warning for https://github.com/GodotVR/godot_openxr/issues/139?

m4gr3d (changes requested):
Just some minor feedback, but it looks mostly good.

Based on the updated logic, I realize I need to update the `start_passthrough` logic as the current implementation doesn't deal well with calling it several times in a row. I'll provide a PR with the update.

m4gr3d:
> Based on the updated logic, I realize I need to update the `start_passthrough` logic as the current implementation doesn't deal well with calling it several times in a row. I'll provide a PR with the update.

Updated the logic in https://github.com/GodotVR/godot_openxr/pull/142

demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
BastiaanOlij: Not yet, it's there for convenience, I'm assuming people who want to support this will want to check if support is available.

demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
BastiaanOlij: Ok, thats a misunderstanding.

These are properties so the values are stored in the TSCN file. After loading Godot will call set_passthrough with the stored value however this is called as part of the loading process when we're not yet ready to call `_start_passthrough`. That is why we call `_start_passthrough` here based on whether the setting was set during load.

While I used to do a syntax here like `set_passthrough(enable_passthrough)`, I've stopped doing this because in Godot 4 the syntax for properties changed and this would result in a never ending loop as setting `enable_passthrough` would call `set_passthrough` again.

demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
BastiaanOlij: We should have access to the version number, I'll need to check.

m4gr3d:
@BastiaanOlij Were you able to update this PR following #142?

demo/addons/godot-openxr/scenes/first_person_controller_vr.gd
BastiaanOlij: Added that in with a few other checks that seemed to make sense.

BastiaanOlij:
@m4gr3d I rebased this project and added a few extra things into it, including being able to switch pass through with a button in the UI that is shown. Works like a charm on my Quest 1 though the demo scene isn't very suitable for pass through, the ground plane definitely gets in the way :)

m4gr3d (approved):

PR #143, opened 2021-12-29
haptic pulse duration was given in microseconds, rather than ms

tom95:
Tested only with the Valve Index controllers on windows. Before the change, I only got a barely perceptible vibration, even when `rumble` was set to 1.

After the change, the `rumble` value has a proper effect again for me.


BastiaanOlij:
Cool! thanks!

BastiaanOlij:
Cool! thanks! If you want please update the `CHANGES.md` document to include a note of this fix, and add you name to the `CONTRIBUTORS.md` file and I'll merge in your change once CI finishes :)

BastiaanOlij (approved):

tom95:
Thanks for the swift reply! I changed the two files accordingly :)

Issue #67, opened 2021-08-05
Android/Quest screen too bright


BastiaanOlij:
Looks like on the Quest using the GLES2 renderer everything becomes too bright. I think this is because OpenXR is expecting colors in linear color space, this may also be a problem if RGBA8 is used on desktop instead of SRGBA8 and that triggering OpenXR to do an additional color correction. 

Need to investigate further.

parubini:
I'm pleased this has been picked up. I am using an Oculus2 and either cable or the new Air Link to Oculus App on a PC. I have  been struggling with the issue of the bright image in the headset for some time now, though thought the issue was more due to my lack of knowledge about how to correctly set the color space. In the configuration I am using, ultimately it would be ideal if the brightness on the PC display (for others watching) matched that of that experienced in the headset.
I hope the PR you refer to will fix this ! Thanks.

BastiaanOlij:
Ok, this is going to be a two parter to fix.

Upstream: https://github.com/godotengine/godot/pull/51780 will now apply `keep_3d_linear` to GLES2 as well and adds logic that both for GLES2 and GLES3 correctly renders the onscreen preview.

Within OpenXR we probably need to stick to using `GL_RGBA8` in which case OpenXR expect the contents to be in linear color space. Interestingly enough we're going from it being waaay to bright, to being slightly darker. It's really weird.

PR #148, opened 2021-12-31
Improve logic for selecting swapchain format to use

BastiaanOlij:
We've been having some weird issues with handling colors in the plugin.

The OpenXR specification is clear, if you use an sRGB texture type it is expecting the rendering to be done in sRGB with the hardware converting to linear where needed, if you use any other texture type the rendering is supposed to be done/remain in linear color space.

Godot renders in linear color space with its GLES3 driver, then color corrects and converts the result to sRGB when copying that into the buffers setup by OpenXR. That works fine.
Godot renders in sRGB color space with its GLES2 driver and when no post effects are used this is rendered directly into the buffers setup by OpenXR.

The problem we encounter is that under a true OpenGLES2 context, sRGB is not supported. On desktop we're actually running OpenGL3 but on Android we run OpenGLES2. We solved this originally by switching to GL_RGBA8 buffers and directing the render engine to render in linear color space, with the GLES2 renderer doing an sRGB -> linear conversion in the shader. The problem here is that too much color data is lost in this conversion resulting in banding and poor results.

This PR changes the logic so that sRGB formats are used whenever possible, and attempts to use a R10B10G10A2 buffer with GLES2 on Android. 

still testing if this all works fine

### DIFF
--- a/src/openxr/OpenXRApi.cpp
+++ b/src/openxr/OpenXRApi.cpp
@@ -9,6 +9,7 @@
 
 #include "openxr/OpenXRApi.h"
 #include "openxr/include/signals_util.h"
+#include "openxr/include/util.h"
 
 #include <algorithm>
 #include <cmath>
@@ -16,6 +17,10 @@
 
 #ifdef ANDROID
 #include <jni/openxr_plugin_wrapper.h>
+#endif
+
+#ifndef GL_FRAMEBUFFER_SRGB_EXT
+#define GL_FRAMEBUFFER_SRGB_EXT 0x8DB9
 #endif
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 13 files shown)


BastiaanOlij:
SteamVR seems to work as advertised
Oculus Desktop XR runtime chooses `GL_SRGB8_ALPHA8` but looks too dark (only tested GLES2)
Oculus Quest runtime doesn't seem to support `GL_RGB10_A2`, it actually has a lot of unknown formats, need to find out what they are...

BastiaanOlij:
Ok, most issues are fixed now, one was a timing issue where we were retrieving get_keep_3d_linear() before our swapchains had been processed and created. Moved this into the callback and it should now be reliable.

The only issue left is figuring out the correct buffer format to use on Quest however it will run correctly with the current settings, just still have some banding. It should run a lot better on desktop. We've tested both with SteamVR and Oculus runtimes on Windows. It should also resolve issues with Oculus Link. 

I do not know if it will work correctly on Linux.

BastiaanOlij:
So further on my previous comment, Malcolm and I spend a bunch of time testing earlier today to see why Oculus Link wasn't working properly with the Quest 2 (Malcolm has a Quest2) and we found out that because we were using an RGBA8 buffer, and RGBA8 isn't available in the desktop Oculus runtime, it was switching to RGBA16 which seems to be not supported by Link.

This PR already fixed that which lead us to the timing issue of when we are checking our keep 3d linear.

That lead me to do a bunch of testing today running Godot natively on the quest and figuring out that this was why the sRGB buffers weren't working before, they actually were working but the setup wasn't being applied properly.

Cleaned things up, did some more testing, and it looks like it is working fine now with `GL_SRGB8_ALPHA8` set as the swapchain format. Rendering can remain in sRGB, OpenXR handles the colors fine, and I'm assuming that this should fix the color banding issue.

m4gr3d:
> OpenXR handles the colors fine

When I test your branch, the end result looks the same as it was on Godot 3.3 (https://github.com/GodotVR/godot_openxr/issues/67), prior to the addition of the `keep_3d_linear` flag. Is that the expected behavior or am I missing some settings configuration?

One way to compare is by running the demo from the `master` branch with `keep_3d_linear` disabled, then run the demo from your branch. When I do that, they look identical.

BastiaanOlij:
@m4gr3d as I already mentioned in chat, it's all too possible that I jumped to conclusions. It seemed to be working correctly on desktop but by the time I finally had things working on Oculus it was fairly late.

This whole sRGB issue is doing my head in. Using RGBA8 which we were doing before, it requires data to be in linear color space but RGBA8 doesn't have enough precision and when it is converted back to sRGB or similar for display we end up with color banding. Using `GL_SRGB8_ALPHA8` is supposed to work around that issue storing data in sRGB and having the hardware do the needed conversions. Obviously it still isn't doing this correctly and it is treating sRGB data asif it is linear.

I will investigate further

src/openxr/OpenXRApi.cpp
```diff
@@ -1512,65 +1513,61 @@ bool OpenXRApi::initialiseSwapChains() {
...
-		*/
-		if (swapchainFormats[i] == GL_RGBA8) {
-			swapchainFormatToUse = swapchainFormats[i];
-			Godot::print("OpenXR Using RGBA swapchain.");
-		}
+	// TODO sRGB mode seems to not work on Quest so we'll leave it disabled until fixed
+	// This does mean that we'll continue to have banding issues in darker colors as 8bits
+	// does not provide enough precision for storing data in linear color space.
+	// requested_swapchain_formats.push_back({ GL_SRGB8_ALPHA8, false });
+	requested_swapchain_formats.push_back({ GL_RGBA8, true });
```
BastiaanOlij, line 1545: @m4gr3d , both Malcolm and I have exhausted all options we can think off for making this work properly. While things work perfect both with SteamVR and Oculus desktop, including as I understand it over link, natively on the Quest `GL_SRGB8_ALPHA8` does not work as advertised forcing us to store data in linear color space which results in the banding issues we've been having.
Equally so, no matter what we try with the `XR_FB_COLOR_SPACE_EXTENSION_NAME`, we don't seem to be able to inform the Quest to treat the data as being in sRGB color space.

It doesn't look like we can use a `R10G10B10A2` buffer format for the swapchain as an alternative, that format would allow for storing color data in linear color space without issue.

As best we can tell this seems to be an issue with the Quest so we can't take it any further and I've commented out the line of code that would allow us to use the `GL_SRGB8_ALPHA8` texture format.

I suggest merging the PR as it is right now because it does fix all the issues for PCVR including using the plugin over airlink with the Quest 2.

We need to get a Meta engineer involved to diagnose this issue on the Quest side of life with this line of code un-commented.

BastiaanOlij:
We probably need someone to confirm if this works properly on Linux as well, both with SteamVR and Monado. @ChristophHaag any chance you have time to take a swing at it?

src/openxr/OpenXRApi.cpp
m4gr3d: I think we should leave `GL_SRGB8_ALPHA8` enabled. I'm noticing that some of the SDK samples have similarly washed out colors so it may just be an issue with the runtime. I'm reaching internally to confirm.

m4gr3d (approved):

BastiaanOlij:
@m4gr3d discussing this further with Malcolm we feel this should be merged as is, the structural changes are an improvement and fix many things, and we can continue to improve from here once we learn more of why the sRGB logic isn't working as expected.

BastiaanOlij:
Ok all, thanks to a suggestion of @m4gr3d I've added an extra line of code that turns off the hardware conversion from linear to sRGB on writing to the framebuffer. This was what was causing things to get over bright as we ended up doubling this conversion.
Hopefully this doesn't have any unexpected side effects. I don't think so because Godot 3 does not rely on this hardware conversion, it only relies on HW sRGB to linear conversion when reading from textures in the GLES3 renderer.

With this change things should look correct color wise inside of the headset on all platforms I've tested (Only Linux missing from this line-up so if someone can test that, it would be good). Obviously not withstanding quality issues related directly to the lower quality settings used on Mobile GPUs (Quest) and the basic downgrade of the GLES2 renderer. But everything should be in the correct color space and we shouldn't have any banding anymore.

The problem that is left is that the preview on screen can be too dark (I've only tested this with the Oculus runtime but I suspect the same applies to SteamVR). This most likely because the XR Compositor is marking the buffer for HW sRGB to linear conversion similar to what we're doing in the GLES3 renderer with textures. Godot is not expecting this conversion. 
Seeing this conversion is in memory using proper floating point variables there is no loss of color precision here (well not until it's written out), but we're not converting it back to sRGB for display.

That last one can be worked around by using a separate viewport for HMD rendering on desktop and outputting the result to preview on screen with a shader that does execute this conversion.

This will require a fix in Godot itself where we either turn off the HW conversion or ensure the conversion back to sRGB is performed in the shader.

I suggest not letting this last issue hold back merging these fixes.

m4gr3d:
Great to hear! Let's ship this!

Malcolmnixon:
This is an analysis made from the https://github.com/GodotVR/godot_openxr/commit/dbc57dc9ea48e9df295b6b2217e6badba0b1378f commit showing the headset brightness is uniform across the different configurations, and the expected darkening of the preview screen:
image

GLES2 running native on the Quest headset still shows the limited color quantization/banding, but I believe this is expected:
image

BastiaanOlij:
@Malcolmnixon it shouldn't have serious banding in the low color range anymore however I believe that the difference in color in the case of your last screenshot has more to do with limitations of the GLES2 renderer then our color conversion issue.

BastiaanOlij:
Ok all, I'm merging this. Tonight/this weekend I plan to prepare the 1.2.0 release including doing a blog post. I have a few things that are drawing my attention today I need to get off my plate first.

Thanks loads for everyone who has been diving into this problem with us, this really was the last show stopping issue preventing people to adopt this plugin.

PR #152, opened 2022-01-11
Add API for accessing tracking confidence

BastiaanOlij:
This PR improves the way we're parsing our location information and uses the `locationFlags` returned to determine how accurate our tracking is.

If `locationFlags` is 0 our confidence will be set to `TRACKING_CONFIDENCE_NONE (0)`.
If our valid bits are set but our tracking bits are not, confidence will be set to `TRACKING_CONFIDENCE_LOW (1)`
If both our set our confidence will be set to `TRACKING_CONFIDENCE_HIGH (2)`

These bits are tracked for both rotation and orientation, if both are valid the both need to be tracked for the confidence to be considered high.

You have access to the confidence directly on the `OpenXRPose` node, for ARVRController nodes we do not expose this so I've added a convenience method to `OpenXRConfig` called `get_tracking_confidence`. It takes the controller id as a parameter (so 1 for left hand, 2 for right hand, same as the ARVRControllers).

### DIFF
--- a/demo/addons/godot-openxr/CHANGES.md
+++ b/demo/addons/godot-openxr/CHANGES.md
@@ -6,6 +6,7 @@ Changes to the Godot OpenXR asset
 - Added interaction profile for the HP Reverb G2 controllers.
 - Removed deprecated `com.samsung.android.vr.application.mode` meta-data tag.
 - Updated repo `README`.
+- Added controller tracking confidence
 
 1.1.1
 -------------------
--- a/src/openxr/OpenXRApi.h
+++ b/src/openxr/OpenXRApi.h
@@ -52,6 +52,12 @@
 // forward declare this
 class OpenXRApi;
 
+enum TrackingConfidence {
+	TRACKING_CONFIDENCE_NONE,
+	TRACKING_CONFIDENCE_LOW,
+	TRACKING_CONFIDENCE_HIGH
+};
+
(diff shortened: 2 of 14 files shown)


demo/addons/godot-openxr/CHANGES.md
```diff
@@ -4,6 +4,7 @@ Changes to the Godot OpenXR asset
 1.1.2
```
m4gr3d, line 4: This should be bumped to `1.2.0`.

src/openxr/OpenXRApi.h
```diff
@@ -52,6 +52,12 @@
 // forward declare this
 class OpenXRApi;
 
+enum TrackingConfidence {
+	TRACKING_CONFIDENCE_NO,
```
m4gr3d, line 56: This should be `TRACKING_CONFIDENCE_NONE` to match the version in the Godot source code.

m4gr3d (changes requested):

demo/addons/godot-openxr/CHANGES.md
BastiaanOlij: Done

src/openxr/OpenXRApi.h
BastiaanOlij: Good point, done!

m4gr3d (approved):

PR #156, opened 2022-01-19
Advance predictive timing on tracked items used *next* frame

BastiaanOlij:
In Godot we update tracking data right before rendering but only the camera position and orientation is immediately used during rendering. It isn't until processing the next frame that all the nodes in Godot get position.

This has the result that any objects that are children of the camera will seem to move as you move your head instead of locked in place, and this has the result that hands seem to lag behind ever so slightly.

OpenXR however is always predicting where elements will be at time of rendering. We have control over this timing and can advance it by one frame when requesting locations.

This PR does so for obtaining the head center (which is requested when we position the camera node), the controller positions and for hand/finger tracking.
When obtaining our view transforms we use the current frames timing.

### DIFF
--- a/demo/addons/godot-openxr/CHANGES.md
+++ b/demo/addons/godot-openxr/CHANGES.md
@@ -7,6 +7,7 @@ Changes to the Godot OpenXR asset
 - Removed deprecated `com.samsung.android.vr.application.mode` meta-data tag.
 - Updated repo `README`.
 - Added controller tracking confidence
+- Use correct predictive timing for controllers.
 
 1.1.1
 -------------------
--- a/src/openxr/OpenXRApi.cpp
+++ b/src/openxr/OpenXRApi.cpp
@@ -2451,6 +2451,23 @@ bool OpenXRApi::check_graphics_requirements_gl(XrSystemId system_id) {
 	return true;
 }
 
+XrTime OpenXRApi::get_next_frame_time() const {
+	if (!initialised || !running) {
+		return 0;
+	}
+
+	// xrWaitFrame not run yet
+	if (frameState.predictedDisplayTime == 0) {
+		return 0;
+	}
+
+	// We retrieve our tracking information right before we render.
+	// We use the current frames predicted display time while rendering.
+	// However when position nodes in our scene, we update this while processing the next frame.
+	// We thus need to advance our frame timing by one frame when retreiving this data.
+	return frameState.predictedDisplayTime + frameState.predictedDisplayPeriod;
+}
+
 XrResult OpenXRApi::acquire_image(int eye) {
 	XrResult result;
 	XrSwapchainImageAcquireInfo swapchainImageAcquireInfo = {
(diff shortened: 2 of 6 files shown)


src/openxr/extensions/xr_ext_hand_tracking_extension_wrapper.cpp
```diff
@@ -194,7 +194,8 @@ void XRExtHandTrackingExtensionWrapper::update_handtracking() {
 		return;
 	}
 
-	const XrTime time = openxr_api->get_frame_state().predictedDisplayTime;
+	const XrTime time = openxr_api->get_frame_state().predictedDisplayTime + openxr_api->get_frame_state().predictedDisplayPeriod; // predict for next frame, that is when we use this info
```
m4gr3d, line 197: Same comment here about updating the variable's name.

It may be more flexible to add a utility method that returns the next frame time and have all three locations call it.
As you mention above, doing so may not be needed for mono applications, so having a centralized location would allow to update the returned `XrTime` value based on the application type.

m4gr3d (changes requested):

demo/addons/godot-openxr/CHANGES.md
```diff
@@ -4,6 +4,7 @@ Changes to the Godot OpenXR asset
 1.1.2
```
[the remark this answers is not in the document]
BastiaanOlij: Oops, forgot about that :) 
Just changed the milestone and made a change on the master branch, once we rebase our PRs it'll follow through.

src/openxr/extensions/xr_ext_hand_tracking_extension_wrapper.cpp
BastiaanOlij: Love the idea for a function so that's what I end up doing, also checked for our mono condition even though there currently are no use cases for it.

src/openxr/OpenXRApi.cpp
```diff
@@ -2896,7 +2914,21 @@ bool OpenXRApi::get_head_center(float world_scale, godot_transform *transform) {
...
+
+	XrTime time;
+	if (form_factor == XR_FORM_FACTOR_HANDHELD_DISPLAY) {
+		// For handheld displays we are rendering MONO and this method is called both for our rendering position
+		// and to position our camera node for the next frame.
+		// It is more important to get the rendering right here.
+		// Note, there currently are no platforms where OpenXR is used with Godot that support this mode
+		// so it's probably a non issue for the time being. This is already resolved in Godot 4.
+		time = frameState.predictedDisplayTime;
+	} else {
```
m4gr3d, line 2926: Would this logic need to be repeated in other locations (e.g: `action.cpp` or `xr_ext_hand_tracking_extension_wrapper.cpp`)? 
If so, you could move this logic in `get_next_frame_time` (would need to be renamed adequately) as well.

m4gr3d (approved):

m4gr3d (approved):

src/openxr/OpenXRApi.cpp
BastiaanOlij: No, this is purely a problem with this function because Godot can call it when it is retrieving the transform with which it renders. That is the only transform that requires the current frame time.

This is only a problem with mono setups like handhelds (which we don't support). With Stereo this method is only called when we update the ARVRCamera nodes position for which we do want the next frames timing. For stereo rendering it calls 'get_view_transform' to get the transform for the individual eyes.

I've already changed the logic for Godot 4, I don't think it's worth worrying about for Godot 3.

PR #161, opened 2022-01-27
Update the README to match the update to the documentation

m4gr3d:


BastiaanOlij (approved):
Love it, best to not duplicate everything so having a to the point readme that points to the documentation makes a lot of sense!

PR #174, opened 2022-02-22
Remove simple controller and move oculus touch forward

BastiaanOlij:
This is a changed as a result of a discussion with Rune, one of the XR engineers at Valve. Seems we're not the only ones with issues here but this change works around it (Unreal has this issue too, Unity doesn't but we think this is because Unity does what we're about to do or something similar).

On SteamVR there is logic that when a controller becomes active that is not defined in the suggested bindings (in this particular test case, a controller for an HTC Cosmos), SteamVR would fall back on the touch controller bindings, and if that doesn't exist, falls back on the simple controller.

More often than not the simple controller was the one being selected, but this definition rarely suffices.

We believe that the most likely cause is that there is a timing issue between Godot (/Unreal) submitting the bindings, and SteamVR consuming them and at the time of processing the controller only the touch definition hasn't been submitted yet.
We can even see this with known controllers as they are bound to the simple controller bindings but then change to a matching binding once that is submitted. Obviously this does not seem to be desired behavior and Valve is looking into whether this is an issue that needs changes on their side, but this PR will provide us with a solution in the meanwhile.

First is that we remove the simple controller definition, there really are very few situations where falling back on this controller definition makes sense, most games won't work with these bindings anyway.
Second is that we make the touch controller the first definition we submit.

### DIFF
--- a/demo/addons/godot-openxr/CHANGES.md
+++ b/demo/addons/godot-openxr/CHANGES.md
@@ -3,7 +3,7 @@ Changes to the Godot OpenXR asset
 
 1.3.0
 -------------------
-- 
+- Removed standard controller
 
 1.2.0
 -------------------
--- a/src/openxr/OpenXRApi.cpp
+++ b/src/openxr/OpenXRApi.cpp
@@ -225,7 +225,7 @@ const char *OpenXRApi::default_action_sets_json = R"===(
 const char *OpenXRApi::default_interaction_profiles_json = R"===(
 [
 	{
-		"path": "/interaction_profiles/khr/simple_controller",
+		"path": "/interaction_profiles/oculus/touch_controller",
 		"bindings": [
 			{
 				"set": "godot",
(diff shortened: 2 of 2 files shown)


m4gr3d:
> First is that we remove the simple controller definition, there really are very few situations where falling back on this controller definition makes sense, most games won't work with these bindings anyway.

Wouldn't switching the order of submission be enough to resolve the issue?
There are potential use-cases where (non-mainstream) OpenXR runtimes may end up relying on the simple controller definition for their own scenarios, especially given that OpenXR is not limited to just games.

BastiaanOlij:
> Wouldn't switching the order of submission be enough to resolve the issue?
> There are potential use-cases where (non-mainstream) OpenXR runtimes may end up relying on the simple controller definition for their own scenarios, especially given that OpenXR is not limited to just games.

The question is, what sort of controllers are you thinking about? There aren't many that really apply here. 
That said, its definitely worth testing out moving it to the end. Though with the cosmos mappings introduced I think we once again cover all major brands anyway.

m4gr3d:
> The question is, what sort of controllers are you thinking about? There aren't many that really apply here. That said, its definitely worth testing out moving it to the end. Though with the cosmos mappings introduced I think we once again cover all major brands anyway.

It's definitely not suited for games, but the profile provides enough component paths to support app interaction (e.g: XR stylus).

BastiaanOlij:
@m4gr3d this is one of my griefs with OpenXR I think, on paper OpenXR should cater for devices you as a developer are unaware off by mapping one of the suggested bindings to new inputs, but in practice so far this promise has hardly been full filled.

SteamVR and Monado are the only platforms that support a multitude of devices and have some of this logic embedded in their runtime but so far each time it seems the only way to really make this work better, is to recompile the plugin against the latest OpenXR source so you can enable the required extension and add in the needed default bindings. That while really the new bindings entry should be enough and shouldn't require compiling against new extensions as they're just strings sent to the runtime.

I hope this improves with time.

That rant over, I'm thinking of closing this PR, it was half a work around and with us now including the Cosmos in our mappings one that has less importance then before.

m4gr3d:
> @m4gr3d this is one of my griefs with OpenXR I think, on paper OpenXR should cater for devices you as a developer are unaware off by mapping one of the suggested bindings to new inputs, but in practice so far this promise has hardly been full filled.
> 
> SteamVR and Monado are the only platforms that support a multitude of devices and have some of this logic embedded in their runtime but so far each time it seems the only way to really make this work better, is to recompile the plugin against the latest OpenXR source so you can enable the required extension and add in the needed default bindings. That while really the new bindings entry should be enough and shouldn't require compiling against new extensions as they're just strings sent to the runtime.
> 
> I hope this improves with time.
> 
> That rant over, I'm thinking of closing this PR, it was half a work around and with us now including the Cosmos in our mappings one that has less importance then before.

@BastiaanOlij Ping on this PR! Should we close it?

BastiaanOlij:
Hey @m4gr3d,

Yeah I think so, with the addition of the other profiles it seems to be selecting the correct one.

Once I'm done with porting and completing the action map GUI for Godot 4 I am planning to experiment a little more with the timing of what we're doing. Discussing with Rune we are preparing the action map right after the session is created but before the session is started. Even though this seems to be correct (and indeed I moved it here after having issues on the Quest) we are theorizing that it is better to perform this action before the session is created. SteamVR may start processing the action map while we're still suggesting bindings and that could also effect what we were seeing here (not to mention a sporadic crash that has all the signs of being a threading/mutex issue).

Issue #173, opened 2022-02-20
predictedDisplayPeriod jumping to ludicrously high fixed value.


CherrySodaPop:
```
frameState.predictedDisplayPeriod: 41666664
frameState.predictedDisplayPeriod: 41666664
frameState.predictedDisplayPeriod: 41666664
frameState.predictedDisplayPeriod: 59652319710029072
ERROR: OpenXR failed to locate space! [XR_ERROR_TIME_INVALID] at: OpenXRApi::xr_result (src\openxr/OpenXRApi.h:343)
frameState.predictedDisplayPeriod: 59652319710029072
ERROR: OpenXR failed to locate space! [XR_ERROR_TIME_INVALID] at: OpenXRApi::xr_result (src\openxr/OpenXRApi.h:343)
frameState.predictedDisplayPeriod: 59652319710029072
frameState.predictedDisplayPeriod: 59652319710029072
ERROR: OpenXR Failed to locate view space in play space! [XR_ERROR_TIME_INVALID] at: OpenXRApi::xr_result (src\openxr/OpenXRApi.h:343)
frameState.predictedDisplayPeriod: 55555552
frameState.predictedDisplayPeriod: 55555552
frameState.predictedDisplayPeriod: 55555552
frameState.predictedDisplayPeriod: 55555552
frameState.predictedDisplayPeriod: 55555552
frameState.predictedDisplayPeriod: 55555552
```
Issue began at this [commit.](https://github.com/GodotVR/godot_openxr/commit/c9b42661d4596e1906edff1fc3528dd9611c22da)

BastiaanOlij:
Hi @CherrySodaPop ,

Can you provide some more information on what system this is on? We're getting the timing information we're using from the XR runtime so knowing which runtime is providing incorrect information will help in finding out what to do about it.

Thanks!

CherrySodaPop:
Heya! I'm using Godot 3.4.2 on Windows 10, the issue only seems to happen on the SteamVR runtime.

BastiaanOlij:
Hi @CherrySodaPop ,

Thanks for that, I'll start investigating and see what I can find, could you also let us know which headset you are using?

CherrySodaPop:
Using the Quest 2 headset with link cable.

BastiaanOlij:
@CherrySodaPop it's a bit of a workaround but can you try #176 ?

CherrySodaPop:
This seems to have done the trick!
[...]

PR #176, opened 2022-02-24
Reset frame state before xrWaitTime and validate display period

BastiaanOlij:
This works around issue #173. It may be that clearing frame state before xrWaitTime is enough, but in case we still get weird numbers, I'm checking for any display period more then 0.5 seconds and clearing it if we receive that. The value retrieved amounted to something like 16570 hours :)

Seeing this is Quest over airlink the best guess I have is that there is some sort of hickup and we're not getting correct data.


BastiaanOlij:
CherrySodaPop mentioned in the original issue that this indeed solves their issue.

PR #177, opened 2022-03-01
Changed info about supported platforms and added shields

BastiaanOlij:
Just some small tweaks to the readme, thinking of adding a screenshot too..


m4gr3d (approved):

BastiaanOlij:
@m4gr3d realized we have a readme in the root and one that will be added to the release. Kept the release one simple and added a bit of detail to the one in the root.

m4gr3d (approved):

PR #178, opened 2022-03-01
Have OpenXR create a depth texture swapchain

BastiaanOlij:
This PR adds support for creating a depth texture swapchain and supplies the correct depth texture to Godot so Godot uses it to render the depth buffer into.

This depth texture is then supplied to OpenXR. OpenXR can use this for advanced reprojection and for tools like LIV. 

This needs thorough testing, on face value it seems to work but as it's a big structural change, best to make sure this works on all platforms first.


src/openxr/OpenXRApi.cpp
```diff
@@ -2968,7 +3079,7 @@ void OpenXRApi::render_openxr(int eye, uint32_t texid, bool has_external_texture
 		return;
 	}
 
-	if (!has_external_texture_support) {
+	if (!has_external_texture_support && !has_external_depth_support) {
```
BastiaanOlij, line 3082: This bit of code has always been here to ensure we have some backwards compatibility with some of the earlier versions of Godot. I'm wondering if we should remove it all together. It would result in the the plugin no longer working on Godot 3.3 and earlier, or whatever version of Godot i added depth texture support to. It's been in there awhile even if we never implemented it in the plugin before.

src/openxr/extensions/xr_fb_foveation_extension_wrapper.cpp
```diff
@@ -52,7 +52,7 @@ void XRFbFoveationExtensionWrapper::on_instance_destroyed() {
 	cleanup();
 }
 
-void **XRFbFoveationExtensionWrapper::set_swapchain_create_info_and_get_next_pointer(void **swapchain_create_info) {
+void **XRFbFoveationExtensionWrapper::set_swapchain_create_info_and_get_next_pointer(void **swapchain_create_info, bool p_depth_buffer) {
```
BastiaanOlij, line 55: @m4gr3d do we need to add the foveation extension when creating the depth buffer or only for the color buffer?

src/openxr/OpenXRApi.cpp
```diff
@@ -2898,27 +2989,54 @@ XrTime OpenXRApi::get_next_frame_time() const {
...
 		.timeout = 17000000, /* timeout in nanoseconds */
 	};
-	result = xrWaitSwapchainImage(swapchains[eye], &swapchainImageWaitInfo);
-	if (!xr_result(result, "failed to wait for swapchain image!")) {
+	result = xrWaitSwapchainImage(color_swapchains[eye], &swapchain_image_wait_info);
+	if (!xr_result(result, "failed to wait for swapchain color image!")) {
 		return result;
 	}
+
+	acquired_image = true;
```
m4gr3d, line 3021: Should this flag be per eye?

src/openxr/OpenXRApi.cpp
m4gr3d: According to our docs, this plugin is only supported starting with Godot 3.4. So if this code only affects Godot 3.3 and earlier, then yes we can remove it.

src/openxr/OpenXRApi.cpp
```diff
@@ -1999,90 +2067,95 @@ bool OpenXRApi::initialiseSwapChains() {
...
 			return false;
 		}
 
-		result = xrEnumerateSwapchainImages(swapchains[i], 0, &swapchainLength[i], nullptr);
-		if (!xr_result(result, "Failed to enumerate swapchains")) {
-			free(swapchainLength);
+		// now create our depth buffer for our view
+		if (!createSwapChain(
+					0,
+					XR_SWAPCHAIN_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT,
```
m4gr3d, line 2118: The Meta's space warp sample uses this flag to create the depth texture swapchain:
`XR_SWAPCHAIN_USAGE_SAMPLED_BIT | XR_SWAPCHAIN_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT`

src/openxr/extensions/xr_fb_foveation_extension_wrapper.cpp
m4gr3d: Based on their documentation and their samples, it looks like it's only needed for the color buffer.
I'll double-check internally.

src/openxr/OpenXRApi.cpp
Wallbraker: For these flags you should only care about what you as app want to use them as, if the runtime needs to sample the textures/images internally then it will add the flags internally.

src/openxr/OpenXRApi.cpp
```diff
@@ -2134,13 +2197,32 @@ bool OpenXRApi::initialiseSwapChains() {
...
+		depth_views[i].subImage.swapchain = depth_swapchains[i];
+		depth_views[i].subImage.imageArrayIndex = 0;
+		depth_views[i].subImage.imageRect.offset.x = 0;
+		depth_views[i].subImage.imageRect.offset.y = 0;
+		depth_views[i].subImage.imageRect.extent.width = render_target_width;
+		depth_views[i].subImage.imageRect.extent.height = render_target_height;
+		depth_views[i].minDepth = 0.0;
+		depth_views[i].maxDepth = 0.0;
+		depth_views[i].nearZ = 0.01;
+		depth_views[i].farZ = 100.0;
```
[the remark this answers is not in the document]
BastiaanOlij: Yup, and set, I'm just initialising the struct with default values here.

src/openxr/OpenXRApi.cpp
BastiaanOlij: No, because we don't start rendering the second eye before the first is finished and by this time the we have released it.

BastiaanOlij:
Ok, with @ChristophHaag fixes added in it looks like its now submitting the depth buffer properly but for me in SteamVR the re-projection screws up, it's like it's trying to reproject into the wrong direction

src/openxr/OpenXRApi.cpp
BastiaanOlij: By the looks of things, Godot 3.2 and earlier. I think I added the logic in 3.3 :)

Ok I cleaned it up, no use holding on to old code.

BastiaanOlij:
Even though this pretty much works and is ready to merge I've changed this back to a draft PR for two reasons:
1) At the moment this is broken on SteamVR, I've added a test for that to disable it but we may want to postpone merging this until we are sure this is indeed the issue and Valve lets us know its fixed
2) I think the way we change our framebuffer in Godot core is causing a performance hit, I'm going to work on a PR to fix that

BastiaanOlij:
Reminder to myself, need to check `XR_KHR_COMPOSITION_LAYER_DEPTH_EXTENSION_NAME` and only add the depth layers if it's supported.

BastiaanOlij:
> Reminder to myself, need to check `XR_KHR_COMPOSITION_LAYER_DEPTH_EXTENSION_NAME` and only add the depth layers if it's supported.

@m4gr3d @ChristophHaag I'm now checking if this extension is available and too my surprise it's not available on Quest? Might only be available on Quest 2? It is available on the Oculus desktop runtime.

BastiaanOlij:
I've put this back to ready for testing, it now properly checks for the extension and I think the code in the plugin itself is fine. Further testing on the performance hit, I think my computer just needed a restart or it may have been the buggy code in SteamVR itself. I can't find any noticable performance difference running with and without the depth buffer logic on when running on the Oculus Runtime.

m4gr3d:
> > Reminder to myself, need to check `XR_KHR_COMPOSITION_LAYER_DEPTH_EXTENSION_NAME` and only add the depth layers if it's supported.
> 
> @m4gr3d @ChristophHaag I'm now checking if this extension is available and too my surprise it's not available on Quest? Might only be available on Quest 2? It is available on the Oculus desktop runtime.

@BastiaanOlij I don't think the Quest (Oculus mobile runtime) supports depth based reprojection. The closest equivalent would be [spacewarp](https://developer.oculus.com/blog/introducing-application-spacewarp/) which has its own [`XR_FB_space_warp`](https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_FB_space_warp) extension.

src/openxr/extensions/xr_fb_foveation_extension_wrapper.cpp
m4gr3d: Yes, it's only needed for the color buffer, so we can return `nullptr` when `p_depth_buffer` is `true`.

BastiaanOlij:
@m4gr3d in due time we'll need to make it so that we have this solution active once we fully support spacewarp though that will likely be a Godot 4 feature as there are no plans that I am aware off of adding motion vectors to Godot 3.

Bit of a shame that they're not sticking to the standard though, I get that they need their own extension to submit the motion vector buffer but it would make sense if they had implemented submitting the depth buffer according to the existing API.

Alas, standards are made to be broken I guess :)

src/openxr/OpenXRApi.cpp
BastiaanOlij: I've added it in, indeed for OpenGL it is probably ignore but it will make it easier once I port the code to Godot 4.

BastiaanOlij:
Ok, I think I've taken this as far as I can for now, we could merge it in just so all the code is there.

For SteamVR we'll remove the check once I get confirmation the issue on their end is resolved. Valve don't use the depth buffer themselves but they do send it on to the Oculus runtime, so it is strange it stops their platform from working properly.

Once we have something for motion vectors we can add spacewarp support (again possibly not until Godot 4) and tie into the depth buffer code here.

PR #179, opened 2022-03-01
Updating to latest godot-xr-tools and adding some more examples

BastiaanOlij:
This PR updates the included godot-xr-tools to the latest version.

I've added the same climbing rack and basket ball thingy that I added to the Godot 4 example. The basketball thingy now keeps score which is new.

image


BastiaanOlij:
Fixed the issues with the ball court game, turns out it was hitting the FPS counter display as that was added to a collision layer. Also added a bounce sound when the basketball hits something with sufficient velocity.

BastiaanOlij:
Note that the remark about class caching in the original post (now removed) only applies for a new project to which godot-xr-tools is added. Once added entries are put into the project file for all the classes. Once they are in the project file it's all good.

PR #186, opened 2022-03-09
Adding HTC tracker support

BastiaanOlij:
Still working on this but this add HTC vive tracker support to the plugin. This is a port of the code I've already added in Godot 4.

Two things still need to happen:
1) there may be a bug in SteamVR itself that needs a fix, waiting on feedback from Valve
2) the code that retrieves the trackers state needs to improve so it doesn't check actions that do not relate to our tracker, right now it presents a bunch of error spam.


BastiaanOlij:
Small update on this, Valve has a fix in beta that resolves the issue with poses not being retrieved correct once the HTC extension is enabled. There are still some issues with detection but will be working on those soon.

BastiaanOlij:
I need to do some testing but now that a new SteamVR is out that has a fix we were waiting for, we should be able to merge this soon.

PR #189, opened 2022-03-15
Clean up swapchains properly with xrDestroySwapchain

id213sin:
The [XrSwapchain](https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XrSwapchain) handle must be eventually freed via the [xrDestroySwapchain](https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#xrDestroySwapchain) function.


BastiaanOlij (approved):
Thanks! Looks good to me!

PR #193, opened 2022-03-21
Fix invalid pose data when controllers are becoming inactive

### LINKED ISSUE #110: Error spam when controllers are off
Getting math error spam when controllers are not turned on (Index controllers on Windows):
```
ERROR: Condition "Math::is_inf(v.x)" is true.
   at: VisualServerScene::instance_set_transform (servers\visual\visual_server_scene.cpp:690)
ERROR: Condition "Math::is_inf(v.x)" is true.
   at: VisualServerScene::instance_set_transform (servers\visual\visual_server_scene.cpp:690)
ERROR: Condition "Math::is_inf(v.x)" is true.
   at: VisualServerScene::instance_set_transform (servers\visual\visual_server_scene.cpp:690)
ERROR: Condition "Math::is_inf(v.x)" is true.
   at: VisualServerScene::instance_set_transform (servers\visual\visual_server_scene.cpp:690)
```

m4gr3d:
Fix #110.


BastiaanOlij:
I'm not a fan of including the controller models into the plugin, they are specific to one platform with no consensus yet by OpenXR on how controller models will be exposed through the OpenXR API. There is also no sure way to pick the correct controller model for the controller actually used. 

IMHO they belong in platform specific libraries or made available as a separate asset. 

We should start looking into that as it seems that Microsoft and Oculus now have two competing APIs for exposing controller models while Valve is I believe working on something as well, we should be able to come up with something that allows us to request a controller model (or scene in this case now that both are GLTF based) and it deciding how this data is loaded depending on which extension is available.

src/openxr/OpenXRApi.cpp
```diff
@@ -3179,7 +3179,11 @@ void OpenXRApi::update_actions() {
 				is_active = default_actions[ACTION_AIM_POSE].action->is_pose_active(input_path);
 			}
 
-			if (is_active) {
+			// Start with our pose, we put our ARVRController on our aim pose (may need to change this to our grip pose...)
+			godot_transform controller_transform;
+			Transform *t = (Transform *)&controller_transform;
+			inputmaps[i].tracking_confidence = default_actions[ACTION_AIM_POSE].action->get_as_pose(input_path, ws, *t);
+			if (is_active && inputmaps[i].tracking_confidence != TRACKING_CONFIDENCE_NONE) {
```
BastiaanOlij, line 3186: I don't think this is the correct fix. If the controller is active we should be checking it, even if tracking is lost. In this case loosing tracking (so say you're holding the controller somewhere the cameras can't see for a bit) would result in the controller being removed. We also end up skipping checking all the other states of the controller which may be valid, just because the controller isn't tracked anymore doesn't mean we're not getting input from the buttons or joystick.

This could lead to gameplay mechanics failing, a player holding a sword suddenly dropping the sword to the ground just because they swing their arm in a way that a controller becomes obscured and looses tracking. I know Michael from VRWorkout ran into these problems and specifically asked us to expose the tracking confidence value so he could react on this properly.

The correct fix in this imho is that if tracking confidence is `TRACKING_CONFIDENCE_NONE`, we simply skip the call to `godot_arvr_set_controller_transform` so the controller remains in its last known position.
We're providing access to the tracking confidence value so people who want to react differently or know that tracking data is unavailable can check this.

In the plugin that is done through an extra entry point on the Pose node and a separate API on the config node, in Godot 4 this is now solved nicer with the confidence state being a part of the core API.

m4gr3d:
> I'm not a fan of including the controller models into the plugin, they are specific to one platform with no consensus yet by OpenXR on how controller models will be exposed through the OpenXR API. There is also no sure way to pick the correct controller model for the controller actually used.
> 
> IMHO they belong in platform specific libraries or made available as a separate asset.
> 
> We should start looking into that as it seems that Microsoft and Oculus now have two competing APIs for exposing controller models while Valve is I believe working on something as well, we should be able to come up with something that allows us to request a controller model (or scene in this case now that both are GLTF based) and it deciding how this data is loaded depending on which extension is available.

The included controller models are just for the demo, not to be used as plugin provided models. I put them in that location because that's where the parent scene is, but they can be moved in the demo only directory.

The bug went on for so long because we didn't have controller models to visualize, hence didn't realize it was happening. I included them in the demo to prevent that from happening.

BastiaanOlij:
> The included controller models are just for the demo, not to be used as plugin provided models. I put them in that location because that's where the parent scene is, but they can be moved in the demo only directory.
> 
> The bug went on for so long because we didn't have controller models to visualize, hence didn't realize it was happening. I included them in the demo to prevent that from happening.

Ah, apologies, I didn't realize the path. Hmm it'll look weird having mismatching controllers when you're on a different platform but I guess it's better then showing nothing.. Indeed not visualizing what is going on does make it easy for bugs to go unnoticed.

src/openxr/OpenXRApi.cpp
m4gr3d: From testing, there's a noticeable period of time (~1 second) where the controllers become inactive (e.g: tracking switches to hands) but `is_active` still returns `true`. In those scenarios, while `is_active` returns `true`, the tracking confidence goes to `TRACKING_CONFIDENCE_NONE`, hence why the logic was structured this way.

Just stopping updating the transform may work to prevent visual glitches (and they're very obvious and annoying), but then we may also be handling invalid input data because if the api behaved as expected, then this fix would not be needed.

We may take the approach of not updating the transform and providing the tracking confidence so the game logic can make its own decisions whether the input data is valid or not. In that scenario, we should provide an `OpenXRController` script that `ARVRController` node can attach that'll quickly provide that data.

BastiaanOlij: > From testing, there's a noticeable period of time (~1 second) where the controllers become inactive (e.g: tracking switches to hands) but `is_active` still returns `true`. In those scenarios, while `is_active` returns `true`, the tracking confidence goes to `TRACKING_CONFIDENCE_NONE`, hence why the logic was structured this way.

Owh that sucks, I can see how that would mess things up.


> We may take the approach of not updating the transform and providing the tracking confidence so the game logic can make its own decisions whether the input data is valid or not. In that scenario, we should provide an `OpenXRController` script that `ARVRController` node can attach that'll quickly provide that data.

That should be easy to do, you could change https://github.com/GodotVR/godot_openxr/blob/master/demo/addons/godot-openxr/scenes/controller.gd#L8 to something like `if get_is_active and OpenXRConfig.get_tracking_confidence(controller_id) != 0:`, we just need a way to obtain a reference to the OpenXRConfig instance created in the main script.

m4gr3d: I've update the fix to only skip setting the transform when the tracking confidence is `NONE`. From testing, that seems to work fine, so I'll omit updating `controller.gd` for the moment.

m4gr3d:
I've addressed the feedback on the PR. It should be ready to go!

BastiaanOlij (approved):

BastiaanOlij:
Thanks!

PR #197, opened 2022-03-29
Add support for the hand tracking aim state extension

m4gr3d:


BastiaanOlij (approved):
This is a great approach I think. As we discussed privately its a bit of a shame Meta introduced a new API call instead of exposing this logic through a new interaction profile, but this mimics that behavior pretty well.

In Godot 4 where an XRController3D node can have multiple poses, we can expose both the aim pose (if available) and the palm pose making it even more generic.

BastiaanOlij:
Don't forget to update the change log :)

PR #201, opened 2022-04-12
Expose the tracking confidence for the tracked hands

m4gr3d:
This also resolves the issue where we were not receiving 'menu' gestures.


BastiaanOlij (approved):
Owh nice! This looks great, I like the improvement to the controller script optionally hiding it if there is no up to date tracking data. This way users can chose whether controllers should remain in their last known position or be hidden.

Merge whenever you're done with any final tweaks :)

PR #208, opened 2022-05-17
Fix swapchain release issue

BastiaanOlij:
Looking into the issue of the swapchain not releasing properly and locking up the left eye buffer when turning on screen recording on the Quest (2). 

Still investigating this but the change so far ensures we end the frame properly even if there was an issue.
Note that we're also including #207 in this just in case this was an issue with an older version of the SDK. 

Waiting on feedback on the Oculus forum to see if anyone else has encountered this:
https://forums.oculusvr.com/t5/OpenXR-Development/Eye-freezes-when-screen-recording/td-p/963673

### DIFF
--- a/src/openxr/OpenXRApi.cpp
+++ b/src/openxr/OpenXRApi.cpp
@@ -2007,6 +2007,12 @@ bool OpenXRApi::initialiseSwapChains() {
 		return false;
 	}
 
+	swapchain_acquired = (bool *)malloc(sizeof(bool) * view_count);
+	if (swapchain_acquired == nullptr) {
+		Godot::print_error("OpenXR Couldn't allocate memory for swap chains", __FUNCTION__, __FILE__, __LINE__);
+		return false;
+	}
+
 	// Damn you microsoft for not supporting this!!
 	// uint32_t swapchainLength[view_count];
 	uint32_t *swapchainLength = (uint32_t *)malloc(sizeof(uint32_t) * view_count);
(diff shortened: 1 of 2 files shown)


mgschwan:
It seems that commit 48b985c makes it harder to cause the issue but it still happened after a few tries

mgschwan:
I think I have a tentative fix for the issue here https://github.com/GodotVR/godot_openxr/commit/484b1a87bd43726620914ba582f590ea189994f7

It uses the changes from @BastiaanOlij and adds a status variable that releases and initializes the swapchains before the next xrBeginFram if they failed to release

I haven't done a thorough test but I had several cases where I forced the issue and it immediately recovered without any noticeable sideeffect

BastiaanOlij:
> It uses the changes from @BastiaanOlij and adds a status variable that releases and initializes the swapchains before the next xrBeginFram if they failed to release

This would definitely work around the problem and if we can't find the root cause it will have to do. That said, I would really like to find the cause of this problem.

mgschwan:
I did do a stress test on this workaround and it seems to work. 
I increased the recording size and bitrate to really tax the Quest and while it has some performance impacts if it happens multiple times in a row it does prevent the eye freeze and does not seem to have a negative impact during normal (not overtaxing the system) operation

BastiaanOlij:
> I did do a stress test on this workaround and it seems to work. I increased the recording size and bitrate to really tax the Quest and while it has some performance impacts if it happens multiple times in a row it does prevent the eye freeze and does not seem to have a negative impact during normal (not overtaxing the system) operation

Well happy to put the workaround in place for the time being, just need to make sure we mark it as a workaround and know why we're doing it so we can remove it once we learn what the proper fix is

BastiaanOlij:
I think its worth merging these changes in. It fixes some of the issues we're having. Do need to PR Michaels workaround fix as well so we can deal with the swapchain getting locked up.

PR #214, opened 2022-06-05
Add workaround for swapchain release error

mgschwan:
Will release and reinitialize all swapchains when an error during release of a swapchain image occurs


src/gdclasses/OpenXRHand.cpp
```diff
@@ -139,8 +144,12 @@ void OpenXRHand::_physics_process(float delta) {
 
 			if (joints[i] != NULL) {
 				if (parents[i] == -1) {
-					// apply our reference frame to our root frame
-					t = reference_frame * t;
+					if (relative_transform) {
+						t = inv_transforms[XR_HAND_JOINT_PALM_EXT] * t;
```
BastiaanOlij, line 148: skipping the reference_frame will give you problems if you've ever called `center_on_hmd`

src/gdclasses/OpenXRHand.cpp
```diff
@@ -19,12 +19,17 @@ void OpenXRHand::_register_methods() {
 	register_method("set_motion_range", &OpenXRHand::set_motion_range);
 	register_property<OpenXRHand, int>("motion_range", &OpenXRHand::set_motion_range, &OpenXRHand::get_motion_range, 0, GODOT_METHOD_RPC_MODE_DISABLED, GODOT_PROPERTY_USAGE_DEFAULT, GODOT_PROPERTY_HINT_ENUM, "Unobstructed,Conform to controller");
 
+	register_method("get_relative_transform", &OpenXRHand::get_relative_transform);
```
BastiaanOlij, line 22: This should be in a separate PR

src/gdclasses/OpenXRHand.cpp
mgschwan: Ah, you are right, I will add it back in and test it

BastiaanOlij:
ping @mgschwan , Could you resolve the merge conflict and apply the requested changes? Then we can merge this in.

BastiaanOlij:
@m4gr3d would like your opinion on this as well, this is a workaround, not really a fix. We have no idea what is causing the actual issue of the swapchain image not being released properly. It seems to be an issue in the Quest itself.

mgschwan:
I've updated my pull request and it says there are no more conflicts.

BastiaanOlij:
@mgschwan I still prefer if the `relative_transform` change is moved into a separate PR. This has nothing to do with fixing the swap chain and I think that one needs further discussion if we want this change.

Also want @m4gr3d feedback on the swapchain fix and see if he is ok with the workaround before we merge.

src/openxr/OpenXRApi.cpp
```diff
@@ -2420,10 +2420,16 @@ bool OpenXRApi::on_state_idle() {
 	for (XRExtensionWrapper *wrapper : registered_extension_wrappers) {
 		wrapper->on_state_idle();
 	}
+
+	emit_plugin_signal(SIGNAL_SESSION_IDLE);
+
```
BastiaanOlij, line 2425: These should also be a separate PR.

src/openxr/OpenXRApi.cpp
```diff
@@ -2420,10 +2420,16 @@ bool OpenXRApi::on_state_idle() {
...
 	return true;
 }
 
 bool OpenXRApi::on_state_ready() {
+
+	emit_plugin_signal(SIGNAL_SESSION_READY);
+
```
BastiaanOlij, line 2432: We are already emitting `SIGNAL_SESSION_BEGUN` on state ready, does it makes sense to have two signals here?

BastiaanOlij:
@mgschwan I've extracted the session and swapchain fixes into separate PRs and I'm merging those. You'll need to redo a PR for the relative transform change.

BastiaanOlij:
Closing as relevant fixes have been applied in other PRs.

Please open a new issue/pr if more is required.

PR #216, opened 2022-06-26
Fix the logic to retrieve and set the play space type

m4gr3d:
Fix for some of the logic errors in https://github.com/GodotVR/godot_openxr/pull/212

### DIFF
--- a/src/gdclasses/OpenXRConfig.cpp
+++ b/src/gdclasses/OpenXRConfig.cpp
@@ -194,46 +194,32 @@ godot::Dictionary OpenXRConfig::get_available_color_spaces() {
 }
 
 int OpenXRConfig::get_play_space_type() const {
-	if (openxr_api == NULL) {
+	if (openxr_api == nullptr) {
 		return XR_REFERENCE_SPACE_TYPE_STAGE;
 	} else {
 		switch (openxr_api->get_play_space_type()) {
 			case XR_REFERENCE_SPACE_TYPE_VIEW:
-				return 0;
 			case XR_REFERENCE_SPACE_TYPE_LOCAL:
-				return 1;
 			case XR_REFERENCE_SPACE_TYPE_STAGE:
-				return 2;
-				//case XR_REFERENCE_SPACE_TYPE_UNBOUNDED_MSFT:
-				//	return ??;
-				//case XR_REFERENCE_SPACE_TYPE_COMBINED_EYE_VARJO:
-				//	return ??;
+				return openxr_api->get_play_space_type();
+
 			default:
-				return 2;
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
I don't get what you're doing here? The reason we have that case is because the enums in OpenXR don't run sequentially. So as new types are added in OpenXR that we want to support, we would return different values for different OpenXR types as the property dropdown enums in Godot have to be sequential. 

Say we want to support `XR_REFERENCE_SPACE_TYPE_COMBINED_EYE_VARJO`, that for us may become enum 3, not 1000121000.

This code is just a prelude to that conversion so that whomever makes the changes for new types we want to support, has a blueprint for adding that conversion.

m4gr3d:
The current logic returns [1 for `LOCAL` and 2 for `STAGE`](https://github.com/GodotVR/godot_openxr/blob/master/src/gdclasses/OpenXRConfig.cpp#L203-L206) but takes in [2 for `LOCAL` and 3 for `STAGE`](https://github.com/GodotVR/godot_openxr/blob/master/src/gdclasses/OpenXRConfig.cpp#L225-L229):
```
// play space is STAGE
int type = get_play_space_type(); // == 2
set_play_space_type(type); // play space is now LOCAL when it should be STAGE
```

BastiaanOlij:
Ah! I see now, what a dumb mistake of mine, so the fix should just be that `case 2` -> `case 1`, and `case 3` -> `case 2` in `set_play_space_type`

Right now its just "luck" that the values match the first 3 entries for the OpenXR constants.

When you look at: https://github.com/GodotVR/godot_openxr/blob/master/src/gdclasses/OpenXRConfig.cpp#L32
```
register_property<OpenXRConfig, int>("play_space_type", &OpenXRConfig::set_play_space_type, &OpenXRConfig::get_play_space_type, 2, GODOT_METHOD_RPC_MODE_DISABLED, GODOT_PROPERTY_USAGE_DEFAULT, GODOT_PROPERTY_HINT_ENUM, "View,Local,Stage"); // we don't support XR_REFERENCE_SPACE_TYPE_UNBOUNDED_MSFT and XR_REFERENCE_SPACE_TYPE_COMBINED_EYE_VARJO at this time.
```

You can see the enum defined for properties. In Godot core we would actually define a proper enum for this and type the return value and parameter for the getter and setter properly, but GDNative is still a bit iffy here.

BastiaanOlij:
I've fixed the issue for now by just addressing the problem so I can get a build out, but feel free to rebase this and we can have a further discussion of whether we want to simplify this so where the values in Godot match those of OpenXR constants we don't do specific cases.

PR #220, opened 2022-07-05
Adding palm pose support

BastiaanOlij:
Adds support for the new Palm pose extension (https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_EXT_palm_pose)

If the extension is not available, we skip registering the bindings. It will be up to the end user to check if the extension is available if and react accordingly.


src/openxr/OpenXRApi.cpp
```diff
@@ -1546,6 +1635,7 @@ bool OpenXRApi::initialiseInstance() {
 
 	// If we have these, we use them, if not we skip related logic..
 	request_extensions[XR_MND_BALL_ON_STICK_EXTENSION_NAME] = &monado_stick_on_ball_ext;
+	request_extensions[XR_EXT_PALM_POSE_EXTENSION_NAME] = &palm_pose_ext;
```
m4gr3d, line 1638: Shouldn't this use our default extension mechanism?

BastiaanOlij: There aren't any functions or anything so it feels like overkill. But otherwise I would say yes. Worth a discussion

BastiaanOlij:
Ok, this is working correctly now. At least with SteamVRs latest beta.

src/openxr/OpenXRApi.cpp
BastiaanOlij: Hmm, thinking some more about this and talking to Rune about the fact that we should only supply bindings that are supported, I think there are a few enhancements we can do here to pretty this up instead of hardcode it. So I will change this.

BastiaanOlij:
Marked this ready for review seeing the functionality is part of OpenXR 1.0.24 and we're dealing with it being an optional feature.

src/openxr/OpenXRApi.h
```diff
@@ -172,6 +172,7 @@ class OpenXRApi {
 
 	// extensions
 	bool monado_stick_on_ball_ext = false;
+	bool palm_pose_ext = false;
```
BastiaanOlij, line 175: This is unused and needs to be removed

src/openxr/OpenXRApi.cpp
```diff
@@ -289,6 +306,14 @@ const char *OpenXRApi::default_interaction_profiles_json = R"===(
 					"/user/hand/right/input/grip/pose"
 				]
 			},
+			{
+				"set": "godot",
+				"action": "palm_pose",
+				"paths": [
+					"/user/hand/left/input/grip/palm_ext",
```
BastiaanOlij, line 313: Change these and others to:
```
"/user/hand/left/input/palm_ext/pose",
"/user/hand/right/input/palm_ext/pose"
```

Issue #234, opened 2022-10-14
Hidden gdns parameters can be lost and cause impossible to fix bugs such as two left hands


goatchurchprime:
Let's take the `right_hand_mesh.tscn` scene which contains the `right_hand.glb` model with the `OpenXRPose.gdns` script attached to it.
![image](https://user-images.githubusercontent.com/677254/195928875-83d9be76-2342-4c99-b5a6-b527783fba65.png)

There are 3 script variables (invisible_if_inactive, action, and path) which can only be seen when the project is running in debug mode in the headset.
![image](https://user-images.githubusercontent.com/677254/195929288-8f142559-411e-45f4-a787-4d50b0f56660.png)

These are registered in the startup phase here:  https://github.com/GodotVR/godot_openxr/blob/master/src/gdclasses/OpenXRPose.cpp#L10 and don't exist otherwise in the editor.  

So how do they get set?  

Well, you can only find them in the scene code when you open it in a text editor.

https://github.com/GodotVR/godot_openxr/blob/master/demo/addons/godot-openxr/scenes/right_hand_mesh.tscn#L18
```
[node name="HandModel" parent="." instance=ExtResource( 2 )]
transform = Transform( 1, 0, 0, 0, -4.37114e-08, -1, 0, 1, -4.37114e-08, 0.5, 1, -0.5 )
script = ExtResource( 3 )
**path = "/user/hand/right"**

[node name="Skeleton" parent="HandModel/Armature" index="0"]
bones/9/bound_children = [ NodePath("IndexTip") ]
script = ExtResource( 1 )
**hand = 1**
```

There is simply no way to see or change these settings in the Godot editor.

What's _really_ bad is that the moment you open the `right_hand_mesh.tscn` in the Godot editor as a new tab, the editor saves over it and deletes all these hidden variables.  This leaves you now with two left hands (the default)!

### What can be done?

Is this problem of easily lost pre-set hidden parameters still in Godot4, or is the system fundamentally different?

The solution that I can think of is to put in a some hacky functions that reset these values, similar to `_update_motion_range()` in `hand_mesh.gd` which sets one of these hidden parameters in the Skeleton (and stops the project running in pancake mode).  

We're going to need similar functions everywhere that there is an OpenXRPose used.

goatchurchprime:
I'll need to redo that pull request from the start again.  I can use `_enter_tree()` to reset the parameters in the lower nodes before their `_ready()` gets called.

PR #235, opened 2022-10-15
Handle missing and accidentally reset OpenXRPose.gdns parameters used in the right hand mesh

goatchurchprime:
This is intended to fix https://github.com/GodotVR/godot_openxr/issues/234

It also sorts out the two instances of gdns parameter accessing which prevent using these nodes in an app on the PC without loading VR, which is sometimes a better place to develop and debug features you don't need to be in a headset to test.

### DIFF
--- a/demo/Main.gd
+++ b/demo/Main.gd
@@ -50,4 +50,5 @@ func _on_FPController_initialised():
 
 func _on_FPController_failed_initialisation():
 	# exit our app
-	get_tree().quit()
+	#get_tree().quit()
+	print("Failed VR initialization, but not quitting!!!")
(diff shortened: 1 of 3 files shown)


goatchurchprime:
Turns out this is not good enough, because when it sets the `OpenXRSkeleton::hand = 1` it is too late because it has already tried and failed to find all the bones (which are ending in '_R' not '_L').

I tried inserting a line:
> $HandModel/Armature/Skeleton.request_ready()
but it didn't work.  (Does it fail with GDNS files?)

The quick work around is the following script on the node `vr_glove_right_slim`:

```GDScript
extends MeshInstance

func _ready():
	get_parent().hand = 1
```

as this gets called before the `OpenXRSkeleton::_ready()` function.

Issue #171, opened 2022-02-20
Need help for pico neo3 device


xvolica:
Pico support using openxr mobile sdk with [guide here](http://developer.pico-interactive.com/document/doc) and [SDK here](https://sdk.picovr.com/developer-platform/sdk/Pico_OpenXR_Mobile_SDK_2.0.1.zip)
So I try to run godot vr demo game on Pico neo3:
```
src/main/cpp/CMakeLists.txt
#set(OPENXR_MOBILE_ROOT_DIR "${CMAKE_SOURCE_DIR}/../thirdparty/oculus_mobile_sdk/OpenXR")
set(OPENXR_MOBILE_ROOT_DIR "${CMAKE_SOURCE_DIR}/../thirdparty/pico_openxr_sdk/")

src/main/AndroidManifest.xml
<meta-data android:name="pvr.app.type" android:value="vr" />
```
But I only got these error log when launching:
```
2022-02-20 11:22:47.292 12311-12311/org.godotengine.godotopenxrdemo.ovr I/Perf: Connecting to perf service.
2022-02-20 11:22:47.301 12311-12311/org.godotengine.godotopenxrdemo.ovr W/ActivityThread: notifyAppLaunchStatus org.godotengine.godotopenxrdemo.ovr org.godotengine.godotopenxrdemo.MainActivity
2022-02-20 11:22:47.336 12311-12311/org.godotengine.godotopenxrdemo.ovr I/GodotPluginRegistry: Initializing Godot plugin OpenXR
2022-02-20 11:22:47.341 12311-12311/org.godotengine.godotopenxrdemo.ovr I/GodotPluginRegistry: Completed initialization for Godot plugin OpenXR
2022-02-20 11:22:47.341 12311-12311/org.godotengine.godotopenxrdemo.ovr W/System.err: java.io.FileNotFoundException: _cl_
2022-02-20 11:22:47.341 12311-12311/org.godotengine.godotopenxrdemo.ovr W/System.err:     at android.content.res.AssetManager.nativeOpenAsset(Native Method)
2022-02-20 11:22:47.341 12311-12311/org.godotengine.godotopenxrdemo.ovr W/System.err:     at android.content.res.AssetManager.open(AssetManager.java:824)
2022-02-20 11:22:47.341 12311-12311/org.godotengine.godotopenxrdemo.ovr W/System.err:     at android.content.res.AssetManager.open(AssetManager.java:801)
[...]

PR #241, opened 2022-12-17
Unified OpenXR loader for Meta and Pico devices

chrox:
This PR is largely ported from `godot4` OpenXR module. And now on Android the `libgodot_openxr.so` will not be linked with the vendor provided proprietary `libopenxr_loader.so`.  And the loader for specific devices will be loaded accordingly at runtime by checking the device brand so that only one APK package is needed for both Meta and Pico devices. 

This PR is tested on Pico 4 and Quest 2 and may help to solve #171.

In order to test on Pico devices, the following meta-data is needed in the AndroidManifest.xml file:
```
<meta-data android:name="pvr.app.type" android:value="vr" />
```


chrox:
The Pico controllers work fine now after adding the interaction profile for `/interaction_profiles/pico/neo3_controller`.

BastiaanOlij:
Sorry @chrox , I totally missed you creating this PR.

I'll need to give it a proper review when I have some time and probably @m4gr3d should have a look at this too.

One problem that I think may end up being a showstopper, and the reason we never added more then Quest support to the plugin, is that deploying loaders from other vendors, even if they are unused, will likely be flagged if you actually want to submit your game to the platforms stores.

Just like with the Godot 4 solution, we would really need a separate plugin for Quest, a separate one for PICO, and a separate one for others, with a choice of which one to deploy on export.
Sadly we don't have that ability in Godot 3.

We might be able to work around that with export filters but it won't be pretty.

I'm hoping with some tweaks coming to OpenGL support in Godot 4 in the next build, we'll be at a point where Godot 4 is a far better choice for XR projects then Godot 3 and we can start advising anyone who want to target Pico or HTC or other Android based XR headsets to port their game over to Godot 4.

chrox:
We have already ported our product to Godot 4 so I'm closing this PR in favor of GodotVR/godot_openxr_vendors#13.

